PR #7, opened 2015-06-08
Added a note about navigating away from ingest tab during arrangement

eckardm:
See this conversation: https://groups.google.com/forum/#!topic/archivematica/8SJBJdvhBXc

Also, you may have to check the RST formatting on the note.


sromkey:
:+1: Thanks Max!

PR #8, opened 2015-09-24
Fix typo in AtoM config section

fiver-watson:
The word Archivematica appears to have been accidentally copied into the middle of the sentence for dip upload configuration as the documenation was updated between versions. I simply removed the accidental text.


sromkey:
Thanks for the catch!

PR #13, opened 2016-09-21
Add notes on metadata-only upload config

sevein:
Feel free to change/correct my grammar, but consider this a reminder that we need to specify in the docs how to set up things on AtoM so the metadata-only DIP upload is possible!


jhsimpson:
:+1:  I think we need this

jhsimpson (approved):

PR #14, opened 2016-11-15
updating instructions for finding manual normalization folder.

jhsimpson:
The path defined in the current instructions assumes an older style development installation.  This pull request updates those instructions to use a path that should be correct in most situations.

### DIFF
--- a/user-manual/ingest/manual-normalization.rst
+++ b/user-manual/ingest/manual-normalization.rst
@@ -71,7 +71,7 @@ normalized according the regular Archivematica workflow.
 
 2. Select "Normalize manually" from the dropdown menu.
 
-3. In your file browser, open /archivematica/src/MCPServer/sharedDirectoryStructure/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/
+3. In your file browser, go to your pipelines Processing Location, and find the approveNormalization watched folder.  The default location is at :file:`/var/archivematica/sharedDirectory/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/'. You may need to look at your storage service configuration to confirm the Processing Location defined for your pipeline.
 
 4. Choose which files you'd like to normalize for preservation and/or access.
 


sromkey:
I think there's a typo, "sahredDirectory" instead of "sharedDirectory". Otherwise looks good to me.

user-manual/ingest/manual-normalization.rst
```diff
@@ -71,7 +71,7 @@ normalized according the regular Archivematica workflow.
 
 2. Select "Normalize manually" from the dropdown menu.
 
-3. In your file browser, open /archivematica/src/MCPServer/sharedDirectoryStructure/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/
+3. In your file browser, go to your pipelines Processing Location, and find the approveNormalization watched folder.  The default location is at /var/archivematica/sahredDirectory/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/ .  You may need to look at your storage service configuration to confirm the Processing Location defined for your pipeline.
```
sevein: Could you use the `:file:` inline markup? e.g.

``` rst
3. In your file browser, go to your pipelines Processing Location, and find the approveNormalization watched folder.  The default location is at :file:`/var/archivematica/sharedDirectory/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/`. You may need to look at your storage service configuration to confirm the Processing Location defined for your pipeline.
```

More info here: http://www.sphinx-doc.org/en/1.4.8/markup/inline.html#role-file. It's handy because it generates different HTML output, e.g.:

``` html
<code class="file docutils literal"><span class="pre">/var/archivematica/sharedDirectory/watchedDirectories/approveNormalization/preservation/yourTransfer123abc/objects/</span></code>
```

And we have pre-defined stylesheets for that kind of stuff.

jhsimpson: ok, will do.

jhsimpson:
Should be fixed now.

sevein:
👍  unless you want to address some nit-picking: 1) we don't use double sentence spacing in our docs and 2) there is also an unnecessary space after  :file:`...` and before the dot.

sevein:
976aced6300a36cdb7667a97f54dca86eb0844e3

PR #15, opened 2017-01-12
Updated Elasticsearch maintenance page for v. 1.6

jrwdunham:
I have updated the maintenance doc page so that it works with AM v. 1.6. In doing so, the following issues were noted:

1. The AM devtool `am rebuild-elasticsearch-transfer-index-from-files` no longer works. I get `ImportError: No module named databaseInterface`. The current inability to rebuild the transfer ES index is stated explicitly in this doc.

2. The query `pyes.TermQuery("mets.ns0:mets_dict_list.ns0:amdSec_dict_list.@ID", "amdsec_1")` seems like not a very good example because every AIP with at least one file will match that query. Can someone suggest a better query example?

### DIFF
--- a/admin-manual/maintenance/maintenance.rst
+++ b/admin-manual/maintenance/maintenance.rst
@@ -49,18 +49,20 @@ data using a Python script that leverages the pyes library.
 
 The first step, when using pyes, is to require the module. The following code
 imports pyes functionality on a system on which Archivematica is installed.
+(Note that you may have to change the path passed to ``sys.path.append`` to
+match your install configuration.)
 
 .. code:: bash
 
    import sys
-   sys.path.append("/home/demo/archivematica/src/archivematicaCommon/lib/externals")
-   from pyes import *
+   sys.path.append("/vagrant/src/archivematica/src/archivematicaCommon/lib/externals")
+   import pyes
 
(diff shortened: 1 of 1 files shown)


Hwesta:
Is this still accurate? Since artefactual/archivematica#138 (pre 1.4?) we have used the official Python elasticsearch library.

jhsimpson:
@Hwesta is right, this PR needsd to be updated to remove the references to pyes, and refer to the official elasticsearch python library https://pypi.python.org/pypi/elasticsearch/1.9.0

Archivematica no longer installs pyes, so these instructions probably won't work as is.

sevein (changes requested):
See comments above.

kellyannewithane:
Fixes #37

sallain:
@sevein Can we close this PR? Is there work here we can re-use?

sevein:
Highly inaccurate! I'd suggest to close, but you may also want to file a new issue to capture the fact that https://www.archivematica.org/en/docs/archivematica-1.8/admin-manual/maintenance/maintenance/#elasticsearch is outdated!

sallain:
Closed in favour of a new issue: https://github.com/archivematica/Issues/issues/292

PR #17, opened 2017-02-28
1.6: Ansible install instructions

Hwesta:
Added instructions for installing Archivematica from source in a production configuration via Ansible.  Formatting is semantic linebreaks (at the end of each sentence) but that can be changed.

A few questions:
* Should the quickstart be updated to install 1.6 instead of qa/1.x?
* Should the quickstart be updated to do a production deploy instead of a development one?
* Or, should the deploy-pub repo be updated to default to 1.6 and/or production? See artefactual/deploy-pub#25

~~A few style tweaks:~~
* ~~Removed leading whitespace that was causing an accidental block format~~
* ~~Removed newlines from lists in quickstart, since that was breaking formatting~~

Edit: since artefactual/deploy-pub#25 was merged, removed the two instructions modifying files in deploy-pub.  Also manually wrapped lines, based on discussions.

### DIFF
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -22,7 +22,7 @@ Installation
 
 * :ref:`Docker <docker>`
 
-* :ref:`Vagrant/Ansible <vagrant>`
+* :ref:`Install for development <development>`
 
 * :ref:`Advanced <advanced>`
 
@@ -631,10 +631,10 @@ Docker
 
 Docker installations are experimental at this time- instructions coming soon.
 
-.. _vagrant:
+.. _development:
 
-Vagrant/Ansible
----------------
+Install for development
+-----------------------
 
 The recommended way to install Archivematica for development is with Ansible and
 Vagrant. For instructions on how to install Archivematica from a virtual machine,
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -294,9 +294,85 @@ If you have trouble with the gearman command try this as an alternative:
 Install from source
 -------------------
 
-Installing from source has been tested using ansible scripts. Ansible installations have been tested for new installations but are not fully tested for upgrades.
+Installing from source has been tested using ansible scripts. Ansible 
+installations have been tested for new installations but are not fully tested 
+for upgrades.
 
-Instructions coming soon.
+These instructions are designed to create a test environment on your local 
+machine. The ansible roles referenced here can be used in production deployments 
+by creating your own ansible playbook to run them. See 
+https://github.com/artefactual/deploy-pub/playbooks/archivematica for more 
+details.
+
+The following instructions detail how to install and run Archivematica from 
+source on a virtual machine.
+
+1. Install VirtualBox, Vagrant, and Ansible:
+
+   .. code:: bash
+
+      sudo apt-get install virtualbox vagrant
+      sudo pip install -U ansible
+
+   Vagrant must be at least version 1.5. Check your version with:
+
+   .. code:: bash
+
+      vagrant --version
+
+   If it is not up to date, you can download the newest version from the 
+   `Vagrant website <https://www.vagrantup.com/downloads.html>`_ .
+
+#. Checkout the deployment repo:
+
+   .. code:: bash
+
+      git clone https://github.com/artefactual/deploy-pub.git
+
+#. Download the Ansible roles:
+
+   .. code:: bash
+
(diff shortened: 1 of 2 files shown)


admin-manual/installation/installation.rst
```diff
@@ -294,9 +294,84 @@ If you have trouble with the gearman command try this as an alternative:
...
+
+#. Set it to install Archivematica 1.6.  Modify the file `singlenode.yml` to say:
+
+   ..code:: yaml
+
+       - include_vars: "vars-singlenode-1.6.yml"
+
+#. Set it to install in production instead of development mode. Modify the file `vars-singlenode.yml` and **remove** the following line:
+
+   ..code:: ruby
```
Hwesta: This and `..code:: yaml` don't seem to format correctly. This doesn't matter if they're removed (see other comment)

admin-manual/installation/installation.rst
```diff
@@ -294,9 +294,84 @@ If you have trouble with the gearman command try this as an alternative:
...
+      cd deploy-pub/playbooks/archivematica
+      ansible-galaxy install -f -p roles/ -r requirements.yml
+
+#. Set it to install Archivematica 1.6.  Modify the file `singlenode.yml` to say:
+
+   ..code:: yaml
+
+       - include_vars: "vars-singlenode-1.6.yml"
+
+#. Set it to install in production instead of development mode. Modify the file `vars-singlenode.yml` and **remove** the following line:
```
Hwesta: This and the previous instruction are not needed if artefactual/deploy-pub#25 is merged.

admin-manual/installation/installation.rst
sevein: The lexer [exists](http://pygments.org/docs/lexers/#pygments.lexers.data.YamlLexer), maybe we have an old version of pygments?

sevein:
I'm sad that we're not wrapping text in archivematica-docs like we've been doing in atom-docs and many big projects out there do too (e.g. https://github.com/django/django/blob/master/docs/index.txt). 😭

sevein:
Should we mention somewhere that Vagrant/VirtualBox are only for local testing/dev purposes? In AtoM we use rst admonitions for this kind of warning.

Hwesta:
Hmm, it looks like if the wrapping in lists is also manually indented, the formatting works (judging by Django's example), so we could update that commit if we wanted to do manual wrapping instead of semantic line breaks.

admin-manual/installation/installation.rst
Hwesta: Possibly! I was only testing the formatting locally, I can't actually confirm if the production version has it. I'd rather not need these instructions if deploy-pub is updated. :)

Hwesta:
> Should we mention somewhere that Vagrant/VirtualBox are only for local testing/dev purposes? In AtoM we use rst admonitions for this kind of warning.

I'm not sure that's true - I think we might support production vagrant installs.  @jhsimpson ?

Hwesta:
Re: Text wrapping, it looks like Kelly fixed the formatting issues I found, so I'm going to rebase this without the formatting changes.

jhsimpson:
VirtualBox is not for production use, at least it would be very much an exception.  Vagrant can work with vmware and other providers, but they way we have set up the deploy-pub repo, to use Vagrant with VirtualBox, would have to be modified, so I think it makes sense to have at least a mention if not a warning that this is not intended for production use.  If for no other reason than the networking would have to change, you can't get to the Vagrant install from outside your local machine without making changes.

admin-manual/installation/installation.rst
```diff
@@ -294,9 +294,82 @@ If you have trouble with the gearman command try this as an alternative:
 Install from source
 -------------------
 
-Installing from source has been tested using ansible scripts. Ansible installations have been tested for new installations but are not fully tested for upgrades.
+Installing from source has been tested using ansible scripts. Ansible 
+installations have been tested for new installations but are not fully tested 
+for upgrades.
+
+While installing using Vagrant is well suited to trying out Archivematica, it is 
+not suitable for production deployment.
```
jhsimpson: suggest changing:
While installing using Vagrant is well suited to trying out Archivematica, it is not suitable for production deployment

to:
These instructions are designed to create a test environment on your local machine.  The ansible roles referenced here can be used in production deployments by creating your own ansible playbook to run them.  See https://github.com/artefactual/deploy-pub/playbooks/archivematica for more details.

jhsimpson:
:+1:

jhsimpson (approved):

PR #20, opened 2017-03-01
Update new install from packages section for 1.6

sbreker:
This change updates the instructions for completing a new AM 1.6 install
from packages on Ubuntu.

### DIFF
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -39,6 +39,8 @@ Archivematica is capable of running on almost any hardware supported by Ubuntu; 
 
 Archivematica can be installed on a single machine, or across many machines to spread the processing workload. See :ref:`Advanced <advanced>`.
 
+Note that Archivematica 1.6.0 has not yet been tested with Elasticsearch 2.x and above.
+
 .. _requirements-small:
 
 Recommended Minimum Requirements for small-scale testing
@@ -194,11 +196,11 @@ a new Storage Service user, into this box and click save.
 
 .. _install-new:
 
-Installing Archivematica 1.5 packages (new install)
+Installing Archivematica 1.6 packages (new install)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Archivematica packages in the past have been hosted on Launchpad, in an Ubuntu PPA (Personal
-Package Archive). With the 1.5.0 release, there is now a new repository at packages.archivematica.org.
+Package Archive). With the 1.6.0 release, there are repositories at packages.archivematica.org.
 This has been introduced to allow one central place to store packages for multiple operating systems.
 
 There are some dependencies still hosted on Launchpad, that have not yet been migrated to packages.archivematica.org.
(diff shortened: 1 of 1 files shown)


admin-manual/installation/installation.rst
```diff
@@ -234,45 +237,71 @@ two different sources of packages.
...
+
+.. code:: bash
+
+   sudo apt-get install elasticsearch
+
+6. Install the storage service package
+
+.. code:: bash
+
+   sudo apt-get install -y --force-yes archivematica-storage-service
```
scollazo: The --force-yes is not needed.

admin-manual/installation/installation.rst
```diff
@@ -234,45 +237,71 @@ two different sources of packages.
...
+
+   wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
+   sudo python /tmp/get-pip.py
+
+8. Install Siegfried
 
 .. code:: bash
 
-   sudo apt-get install -y archivematica-storage-service
+   sudo apt-get install -y --force-yes siegfried
```
scollazo: This shouldn't be needed, siegfried should be installed by mcp-client among other dependencies

scollazo (changes requested):

admin-manual/installation/installation.rst
```diff
@@ -234,45 +237,71 @@ two different sources of packages.
...
+
+.. code:: bash
+
+   sudo ln -s /etc/nginx/sites-available/dashboard.conf /etc/nginx/sites-enabled/dashboard.conf
+
+Start the Elasticsearch service and configure it to start when the system is booted.
+
+.. code:: bash
+
+   sudo /etc/init.d/elasticsearch restart
```
jhsimpson: This could be changed to use upstart instead?

sudo service elasticsearch restart

admin-manual/installation/installation.rst
sevein: I think Steve was using it to work around the signature verification problem. Santi, were the final 1.6 packages signed with [`5236CA08`](https://packages.archivematica.org/1.6.x/key.asc)?

admin-manual/installation/installation.rst
```diff
@@ -234,45 +237,71 @@ two different sources of packages.
...
+   sudo update-rc.d elasticsearch defaults 95 10
```
jhsimpson, line 285: maybe good idea to add a comment to this, describing what this command does?  And possibly move to a new section about enabling services at boot?

admin-manual/installation/installation.rst
sevein: @jhsimpson does that really work? The Elasticsearch package for 14.04 does not make use of upstart - I've tested against v1.7.5.

admin-manual/installation/installation.rst
sbreker: Removed.

admin-manual/installation/installation.rst
sbreker: "11. Start the Elasticsearch service and configure it to start when the system is booted. This step ensures that the Elasticsearch service will start automatically each time the server is rebooted."

admin-manual/installation/installation.rst
scollazo: Yes, there are new packages built from stable/1.6.x branch signed with the production key `5236CA08`

jhsimpson (approved):
:+1: 

looks good

scollazo (approved):

PR #29, opened 2017-07-04
update instructions for 1.6.1 release

jhsimpson:
work in progress - only fresh install on ubuntu updated so far.

### DIFF
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -40,7 +40,7 @@ Technical Requirements
 
 **Operating System**
 
-Archivematica 1.6.1 installation instructions are provided here for the 
+Archivematica |release| installation instructions are provided here for the 
 following operating systems:
 
 * Ubuntu 14.04.5 64 bit Server Edition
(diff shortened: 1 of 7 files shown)


admin-manual/installation/installation.rst
```diff
@@ -40,12 +40,17 @@ Technical Requirements
 
 **Operating System**
 
-Archivematica 1.6.0 has been tested with:
+Archivematica 1.6.1 installation instructions are provided here for the
```
sevein: Instead of `1.6.1` you can use `|release|` which is defined in `conf.py`.

sevein: That would apply too all the rst files.

jhsimpson: I am not sure that it is a good idea to use a variable here.  For example, there is a sentence in this rst file that says:

"Archivematica 1.6.1 requires ElasticSearch 1.x" 

It may not necessarily be valid to change that to say 1.7.0.  The text still has to be reviewed and updated with the new release, to make sure it is accurate.

PR #31, opened 2017-07-27
Add rights.csv to import metadata

joel-simpson:
Update the Import Metadata page to include details of how to import a rights.csv file (as described by: https://wiki.archivematica.org/Object-level_PREMIS_rights_import)
I renamed the "workflow" section to "Importing descriptive metadata with metadata.csv"
and created a new section called "Importing rights metadata with rights.csv"

Please note: 
1) in the wiki page the table showing an example of rights.csv includes only two rows, but the text refers to 3 rows. I've added a 3rd row here to illustrate what i think the text is describing, but this needs a second look by someone who understands the PREMIS mapping better than I do. (I believe it is saying that you can have two "act" sections, so in the row I added I simply changed the value of the "grant_act" column to a value suggested in the PREMIS 2 data dictionary for the act component)

2) I have not made any mention of how this metadata can or should feed into ArchivesSpace. The wiki seems to suggest these details should go here: https://www.archivematica.org/en/docs/archivematica-1.6/user-manual/access/access/#upload-as
(but to be honest I'm not clear from the requirements what this means exactly)

3) I have also noticed that the documentation is also missing details of file_labels.csv (which should be added as an additional section at some point). The file labels functionality is in the "changes for 0.9" section of https://wiki.archivematica.org/METS


jhsimpson (approved):
This looks okay to me.  It might be nice to squash the commits down to one, but not a big deal.

sevein:
608af18d723f2886345b863da7587a1ccb14e1ef and b723d50ae9f0344378dda075b83007abb46b44cb.

PR #47, opened 2017-11-29
Upgrade CentOS instructions for 1.8.x

scollazo:


admin-manual/installation/installation.rst
```diff
@@ -193,8 +193,9 @@ Using 16.04 (Xenial):
 .. code:: bash
 
    sudo wget -O - https://packages.archivematica.org/1.6.x/key.asc | sudo apt-key add -
-   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/1.8.x/ubuntu xenial main" >> /etc/apt/sources.list'
-   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.6.x/ubuntu-externals trusty main" >> /etc/apt/sources.list'
+   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/repos/apt/dev-1.8.x-xenial/ ./" >> /etc/apt/sources.list'
+   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/repos/apt/release-0.11-xenial/ ./" >> /etc/apt/sources.list'
```
sevein: Should we be using `https` instead?

sevein: Also would it be possible to use `packages.archivematica.org` at all times incl. WIP packages? e.g. `https://packages.archivematica.org/1.8.x-dev` or similar?

mamedin:
Hi Santi,

I have found 2 small issues when installing on Centos 7:

**A) Issue 1:**

When running:
```
set -a -e -x
source /etc/sysconfig/archivematica-storage-service
cd /usr/lib/archivematica/storage-service
/usr/share/python/archivematica-storage-service/bin/python manage.py migrate
";
```
I got:
```
+ cd /usr/share/archivematica/storage-service
bash: line 2: cd: /usr/share/archivematica/storage-service: No such file or directory
```
Using:
```
cd /usr/lib/archivematica/storage-service
```
Instead:
```
cd /usr/share/archivematica/storage-service
```
Fixes the issue.

**B) Issue 2:**

When running:

```
set -a -e -x
source /etc/sysconfig/archivematica-dashboard
cd /usr/share/archivematica/dashboard
/usr/share/python/archivematica-dashboard/bin/python manage.py syncdb --noinput
";
```

I got:

```
ValueError: Unable to configure handler u'logfile': [Errno 13] Permission denied: u'/var/log/archivematica/dashboard/dashboard.log'
```

Looking at /var/log/archivematica/dashboard/:

```
[centos@maml-centos storage-service]$ sudo ls -lh /var/log/archivematica/dashboard/
total 0
-rw-rw-r--. 1 root root 0 nov 29 20:20 dashboard.debug.log
-rw-rw-r--. 1 root root 0 nov 29 20:20 dashboard.log
```

After running:
```
[centos@maml-centos storage-service]$ sudo chown archivematica:archivematica /var/log/archivematica/dashboard/*
```

This small issue is fixed.

I have tested with Selinux enforcing mode:

```
[centos@maml-centos ~]$ sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Max kernel policy version:      28
```

mamedin:
I found other 2 small issues on the Ubuntu 16 install:

- The following gpg key is needed by the `fontconfig-config` and `libfontconfig1` packages:

```
sudo wget -O - https://packages.archivematica.org/1.6.x/key.asc | sudo apt-key add -
```

- The `clamav-freshclam` should be restarted instead running `freshclam`:
```
sudo service clamav-freshclam restart
sudo service clamav-daemon start
````
instead:
```
sudo freshclam
sudo service clamav-daemon start
```

File: admin-manual/installation/installation.rst (hunk 1 above)
Comment (scollazo) [reply admin-manual/installation/installation.rst]: The jenkins server doesn't have ssl support for now, but that might change in the near future.

Regarding using the same domain for all packages, it can be done. The original intention was to keep them separate from the official releases to avoid breaking any current install without knowledge.

**Comment by scollazo** [comment]
The permission problem with dashboard logs was fixed [here](https://github.com/artefactual-labs/am-packbuild/commit/6533d40719547f80b7f79dc4a25c265b001cbc64)

File: admin-manual/installation/installation.rst [hunk 2]
```diff
@@ -192,9 +192,11 @@ Using 16.04 (Xenial):
 
 .. code:: bash
 
-   sudo wget -O - https://packages.archivematica.org/1.6.x/key.asc | sudo apt-key add -
-   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/1.8.x/ubuntu xenial main" >> /etc/apt/sources.list'
-   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.6.x/ubuntu-externals trusty main" >> /etc/apt/sources.list'
+   sudo wget -O - http://jenkins-ci.archivematica.org/repos/devel.key | sudo apt-key add -
+   sudo wget -O - https://packages.archivematica.org/1.7.x/key.asc | sudo apt-key add -
```
sevein: Can we move this key somewhere else so it's not under the 1.7.x path? This is going to be a source of confusion.

admin-manual/installation/installation.rst
```diff
@@ -192,9 +192,11 @@ Using 16.04 (Xenial):
...
+   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/repos/apt/dev-1.8.x-xenial/ ./" >> /etc/apt/sources.list'
+   sudo sh -c 'echo "deb [arch=amd64] http://jenkins-ci.archivematica.org/repos/apt/release-0.11-xenial/ ./" >> /etc/apt/sources.list'
+   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.7.x/ubuntu-externals xenial main" >> /etc/apt/sources.list'
```
sevein: Should we create a 1.8.x externals even though if they're just the same today?
That'll save you from having to edit this file again in the future.

admin-manual/installation/installation.rst
```diff
@@ -363,9 +365,19 @@ headless mode)
...
    sudo -u root bash -c 'cat << EOF > /etc/yum.repos.d/archivematica.repo
-   [archivematica]
-   name=archivematica
-   baseurl=https://jenkins-ci.archivematica.org/1.8.x/centos
+   [archivematica-dashboard]
+   name=archivematica-dashboard
+   baseurl=http://jenkins-ci.archivematica.org/repos/rpm/dev-1.8.x
+   gpgcheck=0
+   enabled=1
+   [archivematica-ss]
```
sevein: Cosmetic suggestion, line break after each section ends?

admin-manual/installation/installation.rst
```diff
@@ -405,18 +417,19 @@ Do not enable Elasticsearch if you are running Archivematica in headless mode.
    set -a -e -x
    source /etc/sysconfig/archivematica-storage-service
    cd /usr/share/archivematica/storage-service
-   /usr/lib/python2.7/archivematica/storage-service/bin/python manage.py migrate
-   /usr/lib/python2.7/archivematica/storage-service/bin/python manage.py collectstatic --noinput
+   /usr/lib/python/archivematica-storage-service/bin/python manage.py migrate
    ";
 
-* Now enable and start the archivematica-storage-service and its nginx frontend
+* Now enable and start the archivematica-storage-service, rngd (needed for encrypted spaces) and the nginx frontend
```
sevein: Missing colon at the end of the sentence.

admin-manual/installation/installation.rst
```diff
@@ -446,7 +459,7 @@ Do not enable Elasticsearch if you are running Archivematica in headless mode.
    set -a -e -x
    source /etc/sysconfig/archivematica-dashboard
    cd /usr/share/archivematica/dashboard
-   /usr/lib/python2.7/archivematica/dashboard/bin/python manage.py syncdb --noinput
+   /usr/share/python/archivematica-dashboard/bin/python manage.py syncdb --noinput
```
sevein: I don't think we use syncdb anymore right? We just need to call `migrate`. There are multiple calls to `syncdb` in this document, I think they all need to be changed.

sevein (changes requested):

admin-manual/installation/installation.rst
scollazo: Created a 1.8.x link pointing to 1.7.x, this way we don't need to duplicate efforts when we add a new externals package, and once 1.7.x is released, we can remove the symbolic link and continue as usual.

sevein (approved):
👍

PR #54, opened 2017-12-14
New Dashboard API whitelist mechanism

sevein:
Now if the whitelist is empty the functionality is disabled. This enables the
administrator to disable the whitelist functionality when it is not needed.

### DIFF
--- a/admin-manual/installation/dashboard-config.rst
+++ b/admin-manual/installation/dashboard-config.rst
@@ -444,9 +444,13 @@ you should see the API key.
 IP whitelist
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


sevein:
For the release notes we could have something like the following:

```
Two changes has been made to the API whitelist functionality:

1. The default API whitelist setting is now empty.

2. If the API whitelist setting is empty the user can still authenticate against
the API using the key. The whitelist only kicks in when it lists at least one
IP address.
```

sromkey:
Thanks @sevein I think this looks great. 👍

sromkey (approved):

PR #59, opened 2017-12-28
Add archivesspaceids.csv documentation

jraddaoui:
The matches between AS Archival Objects and SIP files can be done manually
before starting the transfer and be set in a CSV file on the metadata
folder. This commits explains and adds an example of the process.


sromkey:
@ablwr good catch that this isn't merged yet! @jraddaoui it looks good, do you want to go ahead and merge?

jraddaoui:
Done, thanks!

PR #61, opened 2018-01-12
Updates information on importing metadata with a CSV

sallain:
I've updated the sample metadata so that it shows all possible Dublin Core
fields, including the metadata.csv table and the METS output. Note that this
was done for simple objects only; compound objects did not get updated.

I also updated a couple of screenshots.


currmie (approved):
Looks great. I like the lead-in narrative you included just before the sample metadata.csv and METS file snippet. Definitely more user-friendly than just a file dump!

PR #65, opened 2018-01-22
Document DIP Upload and Store configuration options

sallain:
Started out by adding the DIP Upload and DIP Store processing config options, and then it turned into reformatting the whole section. I've added a short description for each processing config and listed the dropdown options themselves, with a sentence fragment that indicates the implication of the choice.

This resolves [this Archivematica issue](https://github.com/artefactual/archivematica/issues/691).


sromkey (commented):
This looks great. In Processing Configuration, would it be worth explaining that at the present time you can only choose an alternate config when doing AIP re-ingest?

sallain:
Good idea @sromkey, I'll add that.

PR #68, opened 2018-02-05
Update install instructions for AM 1.7 (including headless/indexless installations)

jrwdunham:
Most importantly, this instructs users to set the `_SEARCH_ENABLED`-suffixed env vars to `'false'` if they want to run Archivematica "headless", i.e., with Elasticsearch not installed. This commit also includes some formatting and spelling changes.

This also removes the section entitled `Upgrade from Archivematica 1.8.0 to 1.8` because it did not seem to make sense. This section may need to be restored and re-written.

Note that the links to the install/README.md files (e.g., https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPServer/install/README.md) should reference the `_SEARCH_ENABLED` env vars once the headless PR is merged into AM's stable/1.7.x.

I created a Xenial VM and followed these instructions and they work, except that the https://packages.archivematica.org/1.7.x/ type URLs are pointing to 1.6.1 packages, which resulted in the `_SEARCH_ENABLED` env vars having no effect. These instructions need to be re-tested once AM PR 901 is merged and new packages are created for AM 1.7.


admin-manual/installation/installation.rst
```diff
@@ -75,35 +73,42 @@ however, processing large collections will require better hardware. See
...
 * gearman
+* Elasticsearch (optional as of |release|, see below)
 
 Using additional machines requires some additional configuration.
 
 See :ref:`Advanced <advanced>`.
 
-**Notes**:
+.. note:: *Indexing*. Installing Elasticsearch to provide a search index is now
+   optional as of version |release|.  Installing Archivematica in a headless state,
```
scollazo: Why this is called "headless"? We still expect users to use the web interface, afaik.

admin-manual/installation/installation.rst
```diff
@@ -240,14 +255,26 @@ get the most recent version of pip.
 
 8. Install the Archivematica packages
 
+If you are planning running Archivematica in headless mode (i.e., without
+Elasticsearch), then set the relevant environment variables to ``false``. (See
+the `Dashboard install README`_, the `MCPClient install README`_, and the
```
scollazo: Can this be rephrased? The READMEs are pertinent always, not only for installs without ES

scollazo (changes requested):
This might sound dumb, but why installs without ES are being called "headless"?

admin-manual/installation/installation.rst
jrwdunham: You're right. I believe that this "remove Elasticsearch" (a.k.a. "indexless") work is heading in the direction of being able to run Archivematica in a completely "headless" way, such that you could do everything via the API and never even run the dashboard GUI. But clearly we are not there yet, so maybe we should just call this "indexless" until such time as we really have headless functionality.

admin-manual/installation/installation.rst
jrwdunham: Good point. I'll rephrase as you say.

jrwdunham:
@scollazo I've addressed your change requests and updated the indexless env var instructions to work correctly with systemd.

jrwdunham:
@sevein yeah, I agree. I removed the `set-environment` stuff.

admin-manual/installation/installation.rst
```diff
@@ -167,27 +172,60 @@ Archivematica packages are hosted at packages.archivematica.org. This has been
...
+   maintenance updates only mode and will cease to be officially supported in
+   April 2019.
+
 1. Add Archivematica package sources
 
 Add packages.archivematica.org to your list of trusted repositories.
 
 Using 14.04 (Trusty):
 
+Run these three commands right now (**Delete this section when the final release
```
sevein: `s/Delete/delete`?

admin-manual/installation/installation.rst
```diff
@@ -167,27 +172,60 @@ Archivematica packages are hosted at packages.archivematica.org. This has been
...
-   sudo wget -O - https://packages.archivematica.org/1.6.x/key.asc | sudo apt-key add -
-   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.6.x/ubuntu trusty main" >> /etc/apt/sources.list'
-   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.6.x/ubuntu-externals trusty main" >> /etc/apt/sources.list'
+   sudo wget -O - https://packages.archivematica.org/1.7.x/key.asc | sudo apt-key add -
+   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.7.x/ubuntu trusty main" >> /etc/apt/sources.list'
+   sudo sh -c 'echo "deb [arch=amd64] http://packages.archivematica.org/1.7.x/ubuntu-externals trusty main" >> /etc/apt/sources.list'
 
 Using 16.04 (Xenial):
 
+Run these three commands right now (**Delete this section when the final release
```
sevein: `s/Delete/delete`?

sevein (approved):

admin-manual/installation/installation.rst
```diff
@@ -167,27 +172,60 @@ Archivematica packages are hosted at packages.archivematica.org. This has been
...
 Add packages.archivematica.org to your list of trusted repositories.
 
 Using 14.04 (Trusty):
 
+Run these three commands right now (**Delete this section when the final release
+packages are made**):
+
+.. code:: bash
+
+   sudo wget -O - https://packages.archivematica.org/1.7.x/key.asc | sudo apt-key add -
```
scollazo: I think we should add the urls for keys and repos to the conf.py, so we have one central place to do all the changes needed for the release.

admin-manual/installation/installation.rst
```diff
@@ -349,15 +411,18 @@ Some repositories need to be installed in order to fullfill the installation pro
...
    gpgcheck=0
    enabled=1
    EOF'
 
-3. Service depencencies
+3. Service dependencies
 
-Common services like elasticsearch, mariadb and gearmand should be installed
-and enabled before the archivematica install. It can be done with:
+Common services like Elasticsearch, mariadb and gearmand should be installed
```
scollazo: /mariadb/MariaDB/ and gearmand/Gearmand

admin-manual/installation/installation.rst
```diff
@@ -371,13 +436,13 @@ and enabled before the archivematica install. It can be done with:
...
 
-* First, we install the packages:
+* First, install the packages:
 
 .. code:: bash
 
    sudo -u root yum install -y python-pip archivematica-storage-service
 
-* After the package is installed, we need to populate the sqlite database, and
+* After the package is installed, populate the sqlite database, and
```
scollazo: sqlite/SQLite

admin-manual/installation/installation.rst
```diff
@@ -478,19 +533,40 @@ and enabled before the archivematica install. It can be done with:
...
+   sudo -u root sed -i 's/^Example//g' /etc/clamd.d/scan.conf
+
+* Indexless mode:
+
+If you are planning to run Archivematica in indexless mode (i.e., without
+Elasticsearch), then set the relevant environment variables to ``false``
+
+If you are planning on running Archivematica in indexless mode (i.e., without
+Elasticsearch), then modify the relevant systemd EnvironmentFile files by
+adding lines that set the relevant environment variables to ``false``:
```
scollazo: A note like "Remember to stop elasticsearch if you don't want to use it, to free some resources" would be handy

admin-manual/installation/installation.rst
```diff
@@ -753,7 +846,9 @@ better to update the dashboard before updating the mcp components.
 
 7. Disable Unused Services
 
-Archivematica 1.6.0 uses nginx as http server, and gunicorn as wsgi server. This means that some services used in Archivematica 1.5.0 should be stopped and disabled before performing the upgrade.
+Archivematica |release| uses nginx as http server, and gunicorn as wsgi server.
```
scollazo: Nginx and Gunicorn

scollazo (changes requested):
Some minor fixes, sorry for the nitpicking ;)

admin-manual/installation/installation.rst
jrwdunham: @scollazo I think I tried doing that originally and I had some trouble getting sphinx to do that interpolation in code blocks... I'll try again and report back

scollazo: Ok,if it's too hard forget about it. May be ading the full block might make it easier?

admin-manual/installation/installation.rst
```diff
@@ -386,27 +451,31 @@ and enabled before the archivematica install. It can be done with:
...
    source /etc/sysconfig/archivematica-storage-service
-   cd /usr/share/archivematica/storage-service
-   /usr/lib/python2.7/archivematica/storage-service/bin/python manage.py migrate
-   /usr/lib/python2.7/archivematica/storage-service/bin/python manage.py collectstatic --noinput
+   cd /usr/lib/archivematica/storage-service
+   /usr/share/python/archivematica-storage-service/bin/python manage.py migrate
    ";
 
-* And now, we enable and start the archivematica-storage-service and it’s nginx frontend
+* Now enable and start the archivematica-storage-service, rngd (needed for encrypted spaces) and the nginx frontend:
```
scollazo: Rngd?

admin-manual/installation/installation.rst
scollazo: Wsgi too?

admin-manual/installation/installation.rst
jrwdunham: ok, looks like it works if you use `.. parsed-literal::` instead of `.. code:: bash`. As long as we're not relying on code highlighting for bash commands in our docs, I think that'll be ok. See https://stackoverflow.com/questions/25557602/sphinx-substitutions-next-to-angle-brackets/25557867#25557867.

admin-manual/installation/installation.rst
jrwdunham: If the user is deploying an indexless Archivematica, then presumably they will have listened to the notes above this and not installed or started it in the first place. Therefore, I don't think that a note like you describe makes sense here. If you disagree, please let me know.

admin-manual/installation/installation.rst
jrwdunham: I think it's WSGI

admin-manual/installation/installation.rst
jrwdunham: Googling rngd, I don't see any capitalized variants, so I'm gonna leave it lowercase.

jrwdunham:
Closing in favour of https://github.com/artefactual/archivematica-docs/pull/73

PR #72, opened 2018-02-14
Add SELinux instructions

scollazo:
fixes https://github.com/artefactual-labs/am-packbuild/issues/97

### DIFF
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -317,6 +317,20 @@ Update your system
 .. code:: bash
 
    sudo yum update
+   
+If your environment uses SELinux, at a minmum you will need to run the following commands. Additional configuration may be required for your local setup.
+
+.. code:: bash
+
+   # Allow nginx to use ports 8000 and 8001
+   sudo semanage port -m -t http_port_t -p tcp 8000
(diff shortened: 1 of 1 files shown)


jhsimpson:
It is possible to configure SELinux on an ubuntu machine as well, correct?  Shouldn't the SELinux configuration instructions be in a separate section of the docs, and linked to from both ubuntu and centos install sections?

scollazo:
Selinux support for Ubuntu seems pretty outdated:

https://help.ubuntu.com/community/SELinux

Ubuntu uses AppArmor, and I'm not aware of any issues with Archivematica, so nothing to document on that side.

admin-manual/installation/installation.rst
```diff
@@ -317,6 +317,20 @@ Update your system
 .. code:: bash
 
    sudo yum update
+   
+If your environment uses SELinux, you'll need to tweak some configurations:
```
jhsimpson: Suggest adding a sentence after this

`At a minmum you will need to run the following commands.  Additional configuration may be required for your local setup.  `

jhsimpson (approved):
Other than a minor suggested addition, this looks good to me
:+1:

PR #77, opened 2018-02-23
Add instructions for upgrading to indexless

### LINKED ISSUE #74: Problem: upgrading instructions need to be updated
The 1.7 upgrading instructions (admin-manual/installation/installation.rst) current refers to AM 1.5 to 1.6 upgrading. This should be updated to reference 1.6.1 to 1.7.

sallain:
Added a section called "Upgrade to indexless mode" and started to fill out the steps for users who want to upgrade an existing pipeline and stop using Elasticsearch. @jrwdunham there are some code blocks awaiting code that currently "ENTER HERE" - but of course feel free to rework the section as you see fit.

I also did some further style cleanup to bring this page into line with the style guide in the archivematica-docs wiki.

Resolves #74


admin-manual/installation/installation.rst
```diff
@@ -1090,18 +1042,60 @@ This may take a while if you have a large backlog. Once it completes, you
 should be able to see your Transfer Backlog in the Appraisal tab and in the
 Backlog tab.
 
+.. _upgrade-indexless:
 
-.. _advanced:
+Upgrade to indexless mode
```
sallain: @jrwdunham Here's where the upgrade section starts.

PR #83, opened 2018-03-12
Issue 954 update ubuntu firewall instructions

kellyannewithane:
Fixes https://github.com/artefactual/archivematica/issues/954

### DIFF
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -753,7 +753,17 @@ above, follow these steps to complete the configuration of your new server.
    address of the machine you have been installing on).
 
    If you are running the storage service and the dashboard on the same host you
-   should use 'localhost' or '127.0.0.1'
+   should use:
+
+   .. code:: bash
+
+      localhost
+
+   or
+
+   .. code:: bash
+
+      127.0.0.1
 
    If you are using a public IP address you'll need to configure your firewall
    rules and allow access only to port 80 and 8000 for Archivematica usage.
--- a/admin-manual/installation/installation.rst
+++ b/admin-manual/installation/installation.rst
@@ -752,6 +752,12 @@ above, follow these steps to complete the configuration of your new server.
    *test* with the password *test*: http://localhost:8000 (or use the IP
    address of the machine you have been installing on).
 
+   If you are running the storage service and the dashboard on the same host you
+   should use 'localhost' or '127.0.0.1'
+
+   If you are using a public IP address you'll need to configure your firewall
+   rules and allow access only to port 80 and 8000 for Archivematica usage.
+
 2. Create a new administrative user in the Storage Service. The Storage Service
    has its own set of users. In the Users section of the Administration tab of
    the Storage Service, add at least one administrative user, and modify the


admin-manual/installation/installation.rst
```diff
@@ -752,6 +752,12 @@ above, follow these steps to complete the configuration of your new server.
    *test* with the password *test*: http://localhost:8000 (or use the IP
    address of the machine you have been installing on).
 
+   If you are running the storage service and the dashboard on the same host you
+   should use 'localhost' or '127.0.0.1'
```
sevein: The period is missing at the end of the sentence.
Nitpick: `s/storage service/Storage Service` and `s/dashboard/Dashboard`.

You may want to use [inline literals](http://docutils.sourceforge.net/0.6/docs/ref/rst/restructuredtext.html#inline-literals) for things like `localhost` or `127.0.0.1` or the ports below `80` and `8000`.

kellyannewithane:
Fixes #954

admin-manual/installation/installation.rst
```diff
@@ -752,6 +752,22 @@ above, follow these steps to complete the configuration of your new server.
    *test* with the password *test*: http://localhost:8000 (or use the IP
    address of the machine you have been installing on).
 
+   If you are running the storage service and the dashboard on the same host you
```
scollazo, line 755: I think this paragraph is missleading. As the previous one says, you should use the IP address of the machine, localhost will only work when am and ss are being installed in the same computer who will run the browser for AM use.

admin-manual/installation/installation.rst
```diff
@@ -752,6 +752,22 @@ above, follow these steps to complete the configuration of your new server.
...
+      localhost
+
+   or
+
+   .. code:: bash
+
+      127.0.0.1
+
+   If you are using a public IP address you'll need to configure your firewall
+   rules and allow access only to port 80 and 8000 for Archivematica usage.
```
scollazo, line 769: Remove the "only"? Other ports may be needed for management (like 22, ssh), but that's up to the user.

scollazo (commented):
Some feedback

admin-manual/installation/installation.rst
jpellman: Hi- the issue I raised was actually with step 3 of the post-install instructions.  It was actually a bit of confusion on my part- the juxtaposition of "use the IP address of the machine you have been installing on" and "enter the URL of the Storage Service" primed me to use the public IP for the storage service URL, and since there wasn't a rule allowing incoming traffic from the IP range that my public IP was in the dashboard's request to the storage service was blocked.  To clarify- my proposal was that you should make it so that the preferred URL for the Storage Service configuration in step 3 would be localhost/127.0.0.1 if the storage service and dashboard are on the same host.  Sorry if that was a bit unclear.  

The proposed changes to step 1 are actually a bit confusing.  Any discussion of firewall rules should probably happen before the post-install instructions (the CentOS install instructions already do this) and the parts that belong in step 3 don't make sense in the context of step 1.

PR #84, opened 2018-03-13
Add centos development repos

scollazo:


sevein:
Can you remind me why are we doing this? Can't we keep using https://packages.archivematica.org? I find confusing that we have two servers for the same purpose, what's the reason?

admin-manual/installation/installation.rst
```diff
@@ -440,15 +440,22 @@ Archivematica version 1.5.1 and higher support installation on CentOS/Redhat.
...
      .. code:: bash
 
-        sudo -u root bash -c 'cat << EOF > /etc/yum.repos.d/archivematica.repo
-        [archivematica]
-        name=archivematica
-        baseurl=https://packages.archivematica.org/1.7.x/centos
-        gpgcheck=1
-        gpgkey=https://packages.archivematica.org/1.7.x/key.asc
+        sudo -u root bash -c 'cat << EOF > /etc/yum.repos.d/archivematica-dashboard.repo
+        [archivematica-dashboard]
```
sevein, line 444: Why the change from `archivematica` to `archivematica-dashboard`?

scollazo:
Ipackages.a.o is the repository for signed production packages. We don't want to break current installs by adding development packages there. 

Development packages are built by jenkins on each new commit, and are placed in their own repos, one for the ss and one for am, so we need to add both repositories in order to test archivematica.

So , yes, we are using two different urls for repositories with a clear difference: development and production.

This change should be done using sphinx templates, as @jrwdunham did in d785b529646aa6b0 , but  I need to improve my markdown/sphinx skills in order to do a change like that and test it locally.

sevein:
Ok I have a few more questions:

- Why a whole different domain and not something like `https://packages.archivematica.org/1.7.x-devel` or similar to keep things more uniform?
- What is `Ipackages.a.o`?
- I'm still wondering (see my question above) why the repo name was changed from `archivematica` to `archivematica-dashboard`?

> This change should be done using sphinx templates, as @jrwdunham did in d785b52 , but I need to improve my markdown/sphinx skills in order to do a change like that and test it locally.

Ok, I can wait until you figure that out. Maybe the problem is that you're looking at the Markdown reference but Sphinx uses reStructuredText? You can find details about substitution references in the spec: http://docutils.sourceforge.net/0.6/docs/ref/rst/restructuredtext.html#substitution-references.

scollazo:
packages.a.o s was an abbreviation for packages.archivematica.org, sorry about that.

Jenkins runs in a different server than packages.a.o , and uploading packages to the production server  is not configured. I think we can achieve the same effect by doing an internal redirect in packages.archives.org, and avoid configuring jenkins for uploading things.

Have in mind, that from time to time, packages from other branches are also built, and configuring the upload to the production server for each of those custom builds, is a bit of a hassle, as I prefer to have an unique job that builds all packages, using the branch or version as parameters.

The repo name was changed, to avoid using "archivematica", because only the ss or the dashboard were available in each repo. I choose to use names with more significance for the repos (-dashboard for dashboard packages, and -ss for storage-service ones).

Regarding sphinx, I'll check if my changes are compliant with reStructuredText, but will also try to mimic the current production scripts, to be sure that the generated docs are not broken.

sevein:
> The repo name was changed, to avoid using "archivematica", because only the ss or the dashboard were available in each repo. I choose to use names with more significance for the repos (-dashboard for dashboard packages, and -ss for storage-service ones).

I'm still confused. When I go to that repo I still see other packages like `archivematica-mcp-server`.

> We don't want to break current installs by adding development packages there.

I don't see how this is a problem considering that https://packages.archivematica.org/1.7.x/ hasn't been used yet.

---

Can we build packages using the latest tags (`v1.7.0-rc.4` and `v0.11.0-rc.6`) and put them in the canonical repo? Maybe they can be marked as [pre-releases](https://fedoraproject.org/wiki/Package_Versioning_Examples)? I'm insisting on this because I'd rather start testing it now in preparation for the final release.

scollazo:
As proposed by @sevein , rpm packages have been built  with version 1.7.0-rc4 and  0.11.0-rc6, and also have been signed and uploaded to packages.archivematica.org/1.7.x/centos, so this pr can be closed as current docs are up to date.

When final packages arrive, systems installed using an rc version should be able to update without issues, as it it's version number will be greater than the current rc4 one:
<pre>
[root@localhost test]# rpmdev-vercmp 
  Epoch1: 
Version1: 1.7.0
Release1: 1
  Epoch2: 
Version2: 1.7.0
Release2: rc4
1.7.0-1 > 1.7.0-rc4
</pre>

sallain:
@scollazo okay to delete the branch dev/add-centos-development-repos?

PR #104, opened 2018-03-21
sips are created from archival objects but not resources

eckardm:
This is (I think) a slightly less ambiguous way to describe what SIPs can be created from, based on the conversation on the [Appraisal tab Backlog pane / ArchivesSpace pane question](https://groups.google.com/forum/#!topic/archivematica/815b8wk7jR4) thread.

### DIFF
--- a/user-manual/appraisal/appraisal.rst
+++ b/user-manual/appraisal/appraisal.rst
@@ -216,7 +216,9 @@ by the parallel lines icon.
 
 .. note::
 
-   Digital objects created and linked to resources in ArchivesSpace will not appear in the ArchivesSpace panel. However, digital objects added to a resource from within the ArchivesSpace panel are displayed.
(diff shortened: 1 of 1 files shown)


sallain (changes requested):
Max, thanks for this clarification! I think it's good. One slight nitpick, can you edit so that you're truncating after 80 characters, in line with the Archivematica docs style guide? Thanks!

https://github.com/artefactual/archivematica-docs/wiki/Style-guide

eckardm:
Should be all set!

sallain (approved):
Thank you!

sromkey:
This looks good to me, thanks for the fix Max!

sromkey (approved):

sallain:
Merged as 22d9ab9 and b2081d0. Thanks again, Max!

PR #111, opened 2018-03-27
replace drown-down with drop-down

ablwr:
Dark, @sallain, dark... 👻

### DIFF
--- a/user-manual/transfer/forensic.rst
+++ b/user-manual/transfer/forensic.rst
@@ -37,7 +37,7 @@ logs). To make these decisions in the dashboard processing workflow:
 Upload image to dashboard
 -------------------------
 
-1. Select Disk Image type from the drown-down menu on the Transfer tab of the dashboard.
+1. Select Disk Image type from the drop-down menu on the Transfer tab of the dashboard.
 
 2. Add metadata from the imaging process if desired by clicking on the report button.
 


sallain (approved):

ablwr:
@sallain What causes @qubot to close this?

jhsimpson:
Qubot is the GitHub user that is used to push changes from
gitolite.artefactual.com to GitHub.  The repos hosted at
gitolite.artefactual.com are the 'master' repos, and they are where
Artefactual staff push commits.  Qubot then pushes them to GitHub and when
GitHub detects that a PR has been merged, it marks the PR as closed, using
the name of the logged in user.

We have talked at Artefactual about changing this workflow, so that GitHub
becomes the 'master' and committers would push to GitHub instead of
gitolite.  Then PR's would be closed by the person actually merging things
instead of qubot.

I hope that can happen after the 1.7 release.

ablwr:
Ahhh, so someone from Artefactual makes the patch there and then qubot closes it here. Got it! To the average user, it appears like a negation. I can edit the wiki just to note qubot's job in a succinct way, understanding this may change in the future.

ablwr:
I made the change because I have the power but @sallain please adjust if necessary! https://github.com/artefactual/archivematica-docs/wiki/Contributing-guide-for-community-members

PR #125, opened 2018-04-06
Update search index upgrade instructions

### LINKED ISSUE #105: Problem: AM17 upgrade section doesn't mention index update maintenance needs
Reading https://github.com/artefactual/archivematica-devtools/pull/50 I realize that during Archivematica upgrades the operator needs to run certain tools so the search index is updated properly.

I still don't know which tools are exactly needed but the ones that are considered crucial for the upgrade should be converted into management commands (we would be doing this as a separate issue in the archivematica repo) and they should be properly documented which is the purpose of this issue.

The [maintenance page](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/maintenance/maintenance/) lists three of these tools that are part of [archivematica-devtools](https://github.com/artefactual/archivematica-devtools). If rebuilding the AIP/Transfer indices is necessary during upgrades I think that we should state that clearly in the upgrade nots: https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation/installation/#upgrade-from-archivematica-previous-version-x-to-release.

sevein:
The current docs only refer to `rebuild-transfer-backlog`. This commit adds
instructions for an additional `rebuild-elasticsearch-aip-index-from-files`
tool that needs to be executed during upgrades and it also refers to the new
Django management commands instead of `devtools`.

This closes #105.

### DIFF
--- a/admin-manual/installation-setup/upgrading/upgrading.rst
+++ b/admin-manual/installation-setup/upgrading/upgrading.rst
@@ -9,9 +9,11 @@ Upgrade from Archivematica |previous_version|.x to |release|
 * :ref:`Create a backup <create-backup>`
 * :ref:`Upgrade Ubuntu package install <upgrade-ubuntu>`
 * :ref:`Upgrade CentOS/Red Hat package install <upgrade-centos>`
+* :ref:`Upgrade search indices <upgrade-search-indices>`
 * :ref:`Upgrade in indexless mode <upgrade-indexless>`
 
 .. note::
+
    While it is possible to upgrade a GitHub-based source install using ansible,
    these instructions do not cover that scenario.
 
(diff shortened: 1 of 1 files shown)


admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -273,53 +212,57 @@ Upgrade from Archivematica |previous_version| for CentOS/Red Hat
...
-   unless you have a back-up you can restore from.
+Upgrade search indices
+----------------------
 
-1. Install devtools
+.. note::
 
-   .. code:: bash
+   Ignore this section if you are not planning to use the searching
+   capabilities of Archivematica. Instead, follow the instructions on `how to
```
sromkey: Maybe instead, "Ignore this section if you are planning to run Archivematica without a search index."

sromkey (approved):
I had one small re-wording suggestion but otherwise this looks great to me @sevein !

admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -273,53 +212,57 @@ Upgrade from Archivematica |previous_version| for CentOS/Red Hat
...
+   ``command-shift-F5``.
+
+2. **Rebuild the AIP index:**
+
+   Run the following command passing the path of the AIP storage location you
+   confirmed above:
+
+   .. code:: bash
+
+      /usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/python /usr/share/archivematica/dashboard/manage.py rebuild_elasticsearch_aip_index_from_files /var/archivematica/sharedDirectory/www/AIPsStore
```
sevein: @mamedin mentioned that the environment strings need to be loaded, e.g.:

    export $(grep -v ^# /etc/sysconfig/archivematica-dashboard)

Or we could:

```
    set -a -e -x
    source /etc/sysconfig/archivematica-dashboard
    cd /usr/share/archivematica/dashboard
    /usr/share/archivematica/virtualenvs/archivematica-storage-service/bin/python manage.py migrate
";
```

This could be notably simplified by https://github.com/artefactual-labs/am-packbuild/pull/160 but I don't think that's going to happen in AM17.

admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -273,53 +212,57 @@ Upgrade from Archivematica |previous_version| for CentOS/Red Hat
...
 
-3. Using the path you confirmed above, replace the text '/path/to/transfers'
-   with the correct path for your system.
+   Run the following command passing the path of the transfer backlog location
+   you confirmed above:
 
    .. code:: bash
 
-       am rebuild-transfer-backlog /path/to/transfers
+      /usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/python /usr/share/archivematica/dashboard/manage.py rebuild_transfer_backlog /var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog
```
sevein: @mamedin mentioned that the environment strings need to be loaded, e.g.:

    export $(grep -v ^# /etc/sysconfig/archivematica-dashboard)

Issue #121, opened 2018-04-04
Another Wrong Command in CentOS Installation Instructions


jpellman:
Version of the documentation: 1.7

Page (and section, if applicable) where the issue occurs: [CentOS Installation Instructions](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation-setup/installation/install-centos/#install-pkg-centos), step 5, third bullet point

Description of the issue:

[Previous iterations of the 1.7 docs](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation/installation/#install-pkg-centos) have included the following command for this step:

```
sudo -u archivematica bash -c " \
set -a -e -x
source /etc/sysconfig/archivematica-dashboard
cd /usr/share/archivematica/dashboard
/usr/share/python/archivematica-dashboard/bin/python manage.py migrate
";
```

The current iteration (post re-organization of the docs) includes the following command for this step:

```
sudo -u archivematica bash -c " \
set -a -e -x
source /etc/sysconfig/archivematica-dashboard
cd /usr/share/archivematica/dashboard
/usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/python manage.py migrate
";
```

Which results in the following error:

```
[centos@ip-172-31-10-71 ~]$ sudo -u archivematica bash -c " \
> set -a -e -x
> source /etc/sysconfig/archivematica-dashboard
> cd /usr/share/archivematica/dashboard
> /usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/python manage.py migrate
> ";
+ source /etc/sysconfig/archivematica-dashboard
++ ARCHIVEMATICA_DASHBOARD_DASHBOARD_DJANGO_SECRET_KEY=pZf2I02qNiNH7TyUzzPQCDv4lKZsi30F
++ ARCHIVEMATICA_DASHBOARD_DASHBOARD_DJANGO_ALLOWED_HOSTS='*'
++ AM_GUNICORN_BIND=127.0.0.1:7400
++ DJANGO_SETTINGS_MODULE=settings.production
++ ARCHIVEMATICA_DASHBOARD_DB_NAME=MCP
++ ARCHIVEMATICA_DASHBOARD_DB_USER=archivematica
++ ARCHIVEMATICA_DASHBOARD_DB_PASSWORD=demo
++ ARCHIVEMATICA_DASHBOARD_DB_HOST=localhost
[...]

Issue #120, opened 2018-04-04
Wrong Command in CentOS 7 Installation Instructions


jpellman:
Version of the documentation: 1.7

Page (and section, if applicable) where the issue occurs: [CentOS Installation Instructions](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation-setup/installation/install-centos/#install-pkg-centos), step 4, second bullet point

Description of the issue:

[Previous iterations of the 1.7 docs](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation/installation/#install-pkg-centos) have included the following command for this step:

```
sudo -u archivematica bash -c " \
set -a -e -x
source /etc/sysconfig/archivematica-storage-service
cd /usr/lib/archivematica/storage-service
/usr/share/python/archivematica-storage-service/bin/python manage.py migrate
";
```

The current iteration (post re-organization of the docs) includes the following command for this step:

```
sudo -u archivematica bash -c " \
set -a -e -x
source /etc/sysconfig/archivematica-storage-service
cd /usr/lib/archivematica/storage-service
/usr/share/archivematica/virtualenvs/archivematica-storage-service/bin/python manage.py migrate
";
```

Which results in the following error:

```
[centos@ip-172-31-10-71 ~]$ sudo -u archivematica bash -c " \
> set -a -e -x
> source /etc/sysconfig/archivematica-storage-service
> cd /usr/lib/archivematica/storage-service
> /usr/share/archivematica/virtualenvs/archivematica-storage-service/bin/python manage.py migrate
> ";
+ source /etc/sysconfig/archivematica-storage-service
++ DJANGO_SECRET_KEY=CHANGE_ME_WITH_A_SECRET_KEY
++ DJANGO_STATIC_ROOT=/usr/lib/archivematica/storage-service/assets
++ DJANGO_SETTINGS_MODULE=storage_service.settings.production
++ EMAIL_HOST=localhost
++ EMAIL_HOST_PASSWORD=
++ EMAIL_HOST_USER=
++ EMAIL_PORT=25
++ SS_DB_HOST=
++ SS_DB_NAME=/var/archivematica/storage-service/storage.db
++ SS_DB_PASSWORD=
++ SS_DB_USER=
++ SS_DB_NAME=/var/archivematica/storage-service/storage.db
++ DJANGO_ALLOWED_HOSTS='*'
++ SS_GUNICORN_BIND=127.0.0.1:7500
[...]

PR #126, opened 2018-04-10
Use CentOS/Ubuntu dev repos during dev stage

sevein:
We were already asking users to install the temporary repos in Ubuntu environments. This just extends the same solution to CentOS environments.

This closes https://github.com/artefactual/archivematica-docs/issues/121 and https://github.com/artefactual/archivematica-docs/issues/120.

### DIFF
--- a/admin-manual/installation-setup/installation/install-centos.rst
+++ b/admin-manual/installation-setup/installation/install-centos.rst
@@ -68,6 +68,36 @@ Installation instructions
 
    * Archivematica:
 
+   While Archivematica 1.7 is in development, please use these commands to
+   install the development repositories:
+
+   .. code:: bash
+
+      sudo -u root bash -c 'cat << EOF > /etc/yum.repos.d/archivematica.repo
+      [archivematica]
+      name=archivematica
+      baseurl=http://jenkins-ci.archivematica.org/repos/rpm/release-1.7/
+      gpgcheck=0
+      enabled=1
+      [archivematica-storage-service]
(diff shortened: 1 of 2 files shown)


admin-manual/installation-setup/installation/install-centos.rst
```diff
@@ -68,6 +68,36 @@ Installation instructions
 
    * Archivematica:
 
+   Use the following commands to install the repositories that we are using
```
sallain: Would suggest making this a bit more explicit:

"While Archivematica 1.7 is in development, please use these commands to install the development repositories:"

admin-manual/installation-setup/installation/install-centos.rst
```diff
@@ -68,6 +68,36 @@ Installation instructions
...
+      name=archivematica-extras
+      baseurl=https://packages.archivematica.org/1.7.x/centos-extras
+      gpgcheck=1
+      gpgkey=https://packages.archivematica.org/1.7.x/devel.key
+      enabled=1
+      EOF'
+
+   Use the following commands instead to install the final repositories. This is
+   only relevant once we complete the release.
+
```
sallain: "If the release has been completed, you should use these commands to install the final repositories:"

sallain (commented):
Made a couple suggestions to really make it clear which commands should be used. After the release, I'd like to see the first section removed so that there's no confusion.

sevein:
Thank you, fixed!

sallain (approved):

PR #131, opened 2018-04-16
Issue: IP requirements are unclear

### LINKED ISSUE #129: IPs need not be public in post-install instructions
Version of the documentation: 1.7

Page (and section, if applicable) where the issue occurs: Steps 1 and 5 of post-install instructions [here](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation-setup/installation/install-centos/#centos-post-install-config) and [here](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation-setup/installation/install-ubuntu/#ubuntu-post-install-config).

Description of the issue:

My phrasing in https://github.com/artefactual/archivematica/issues/954 was a bit confusing:

```
indicate that if you are going to use a public IP (i.e., if the storage service and the dashboard are running on separate nodes), you'll need to configure firewall rules (iptables, firewalld, AWS security groups, etc).
```

In this phrase I put `public IP` just because that was what I was working with at the time.  The essential bit was that if services are on separate nodes, firewall-like services need to be configured so that the nodes can talk with each other.  The crux of what I was getting at applies to IPs in both private and public ranges; also there's no real necessity to use IPs if you have domain names set up.  It might arguably make even more sense to use a domain name for a non-standard setup in the future where requests to the storage service are load-balanced across multiple hosts (not something you'd do with these instructions; for standard instructions I wouldn't worry about this).

In this sense step 1 would more accurately read _"If you are using an IP address or fully-qualified domain name instead of localhost"_ instead of _"If you are using a public IP address"_, and step 5 would more accurately read _"you should use the IP address or fully-qualified domain name of your Storage Service instance"_ instead of _"you must use the public IP of your Storage Service instance"_.

sallain:
This clarifies the IP requirements for post-install configuration (centos, ubuntu,
and ansible), changing "public IP" to the more precise "an IP address or fully-
qualified domain name instead of local host."

Resolves #129.

### DIFF
--- a/admin-manual/installation-setup/installation/install-ubuntu.rst
+++ b/admin-manual/installation-setup/installation/install-ubuntu.rst
@@ -388,9 +388,9 @@ the configuration of your new server.
    ``http://127.0.0.1:8000`` if this is a local development setup). The default
    username and password are ``test``/ ``test``.
 
-   If you are using a public IP address, you will need to configure your
-   firewall rules and allow access only to ports 80 and 8000 for Archivematica
-   usage.
+   If you are using an IP address or fully-qualified domain name instead of
+   localhost, you will need to configure your firewall rules and allow access
+   only to ports 81 and 8001 for Archivematica usage.
 
 2. The Storage Service has its own set of users. Navigate to
    **Administration > Users** and add at least one administrative user. After
@@ -422,11 +422,12 @@ the configuration of your new server.
      the same machine, then you should supply ``http://127.0.0.1:8000`` as the
      Storage Service URL at this screen.
    - If the Storage Service and the Archivematica dashboard are installed on
-     different nodes (servers), then you must use the public IP of your Storage
-     Service instance, e.g., ``http://<MY-IP-ADDR>:8000`` *and* you must ensure
-     that any firewall rules (i.e., iptables, ufw, AWS security groups, etc.)
-     are configured to allow requests from your dashboard IP to your Storage
-     Service IP on the appropriate port.
+     different nodes (servers), then you should use the IP address or
+     fully-qualified domain name of your Storage Service instance,
+     e.g., ``http://<MY-IP-ADDR>:8000`` *and* you must ensure that any firewall
+     rules (i.e., iptables, ufw, AWS security groups, etc.) are configured to
+     allow requests from your dashboard IP to your Storage Service IP on the
+     appropriate port.
 
 :ref:`Back to the top <install-pkg-ubuntu>`
 
--- a/admin-manual/installation-setup/installation/install-ubuntu.rst
+++ b/admin-manual/installation-setup/installation/install-ubuntu.rst
@@ -390,7 +390,7 @@ the configuration of your new server.
 
    If you are using an IP address or fully-qualified domain name instead of
    localhost, you will need to configure your firewall rules and allow access
-   only to ports 81 and 8001 for Archivematica usage.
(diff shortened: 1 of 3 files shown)


sevein:
CC @hakamine @mamedin @scollazo

admin-manual/installation-setup/installation/install-ubuntu.rst
```diff
@@ -388,9 +388,9 @@ the configuration of your new server.
    ``http://127.0.0.1:8000`` if this is a local development setup). The default
    username and password are ``test``/ ``test``.
 
-   If you are using a public IP address, you will need to configure your
-   firewall rules and allow access only to ports 80 and 8000 for Archivematica
-   usage.
+   If you are using an IP address or fully-qualified domain name instead of
+   localhost, you will need to configure your firewall rules and allow access
```
mamedin, line 392: I think it is a typo, it should be: ports 80 and 8000

mamedin (changes requested):

admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -273,53 +212,71 @@ Upgrade from Archivematica |previous_version| for CentOS/Red Hat
...
+2. **Rebuild the AIP index:**
+
+   Run the following command passing the path of the AIP storage location you
+   confirmed above:
+
+   .. code:: bash
+
+      sudo -u archivematica bash -c " \
+          set -a -e -x
+          source /etc/sysconfig/archivematica-dashboard
```
mamedin: The`/etc/sysconfig/archivematica-dashboard` config file is only for CentOS. The Ubuntu's file is `/etc/default/archivematica-dashboard`. I think this should be indicated in a note or explicitly for each linux distribution.

admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -273,53 +212,71 @@ Upgrade from Archivematica |previous_version| for CentOS/Red Hat
...
-   with the correct path for your system.
+   Run the following command passing the path of the transfer backlog location
+   you confirmed above:
 
    .. code:: bash
 
-       am rebuild-transfer-backlog /path/to/transfers
+      sudo -u archivematica bash -c " \
+          set -a -e -x
+          source /etc/sysconfig/archivematica-dashboard
```
mamedin: The/etc/sysconfig/archivematica-dashboard config file is only for CentOS. The Ubuntu's file is /etc/default/archivematica-dashboard. I think this should be indicated in a note or explicitly for each linux distribution.

mamedin (changes requested):

admin-manual/installation-setup/upgrading/upgrading.rst
sallain: @mamedin can you start a new issue?

admin-manual/installation-setup/installation/install-ubuntu.rst
sallain: Good catch! Resolved.

admin-manual/installation-setup/upgrading/upgrading.rst
mamedin: @sallain done: https://github.com/artefactual/archivematica-docs/issues/139

mamedin (approved):

Issue #92, opened 2018-03-16
CentOS 7 Ports Not Used in Post-Install Instructions


jpellman:
In the [post-install instructions of Archivematica 1.7](https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation/installation/#post-install-configuration), step 1 gives the wrong port number for a CentOS installation (8000 instead of 8001) and step 3 doesn't give the dashboard port (81).  There should be a note or two indicating the correct ports to use for a CentOS 7 install.

jpellman:
For consistency and clarity, could step 3 in the post-install instructions include language similar to step 1?  So something like:

```
To finish the installation, use your web browser to navigate to the Archivematica dashboard using the IP address of the machine you have been installing on, e.g., http://<MY-IP-ADDR>:81 (or http://localhost:81 or http://127.0.0.1:81 if this is a local development setup).
```

kellyannewithane:
Whoops!  I accidentally pushed my changes directly.  Made changes in centOS, Ubuntu and Ansible.

jpellman:
Hi- this change shouldn't actually apply to Ubuntu, just CentOS, so it just needs to be applied to the CentOS page.

sallain:
@jpellman The post-install configuration instructions are identical regardless of environment (barring the difference in ports), and the amendment to clarify the IP address is a generic suggestion that's not tied to CentOS specifically.

It has been applied to the CentOS installation page here: https://www.archivematica.org/en/docs/archivematica-1.7/admin-manual/installation-setup/installation/install-centos/#post-install-configuration

PR #140, opened 2018-04-24
Post-install clarification at step 3

kellyannewithane:
Fixes https://github.com/artefactual/archivematica-docs/issues/92


sallain (changes requested):
This change should also be reflected on Ubuntu and Ansible pages so that the instructions are internally consistent (`admin-manual/installation-setup/installation/install-ansible.rst` and `admin-manual/installation-setup/installation/install-ubuntu.rst`)

sallain:
Merged directly, rather than through PR.

PR #147, opened 2018-04-28
Fix trusty service instructions

sevein:
systemctl is not available in trusty.

fits is already enabled, the command is not needed.


mamedin (approved):
LGTM

PR #151, opened 2018-05-01
note raw format specification

ablwr:
Considered using "optimally" or "reliably" or some other softener because other formats MAY be possible, especially if user is willing to modify and customize their FPR, but "out of the box" only RAW will work reliably. Suggestions?


sallain (approved):

sallain:
Because this section is intended to be step-by-step instructions on basic forensic procedures, I'd leave the ambiguity out. Maybe in the FPR docs there is somewhere that you could describe modifying commands for disk images - if so, that would be a good thing to hyperlink from the forensics page.

ablwr:
I thought this was merged in but I don't see it, did it get lost?

sallain:
@ablwr sorry I didn't see this earlier - I think it got lost. Want to try re-doing it for the 1.8 docs?

ablwr:
Thanks, I will! This is importantttttt

ablwr:
I need to do a little more than just recreate this because I think it could be clearer, but I will keep this on my list of things to do.

Issue #86, opened 2018-03-14
Problem: FPR docs aren't in main docs repo


sallain:
The Format Policy Registry documentation is located on a branch in the [archivematica-fpr-admin](https://github.com/artefactual/archivematica-fpr-admin) repo. It's hard to remember where it is, and it gets neglected. As it's just a single page, it would be trivial to move the FPR docs to the main documentation repo. This makes sense especially in light of [Archivematica #970](https://github.com/artefactual/archivematica/issues/970), Problem: FPR server should be decoupled from Archivematica.

sromkey:
I agree- @sallain, since I’m working on updating the Pres Planning docs anyway do you want me to take this on and combine the two?

sallain:
@sromkey Go for it!

sallain:
@romkey this got done, right? Are there any remaining tasks to complete?

romkey:
@sallain I don't think this is the @romkey you're looking for :-)

sallain:
@romkey It is not! This is the issue with different Slack names vs Github names 😂 

@sromkey, see above!

sromkey:
Haha, @romkey nice to meet you. Maybe we're long lost relatives! ;)

@sallain, not quite resolved. We resolved it in the 1.7 docs but need a PR from 1.6.1 and earlier because the situation with the FPR server is slightly different in those versions (not decoupled, only turned off).

sallain:
@sromkey I've added the admonition back to 1.4, which is the earliest version of the docs available on the website. I don't think it's worth bothering with anything earlier. I didn't add anything to the FPR docs (https://github.com/artefactual/archivematica-fpr-server/tree/master/docs), as I believe that whole repo will be deleted eventually - let me know if this is incorrect.

sromkey:
That sounds good. I don't know if the repo will be deleted to archived or what, but definitely no need to update them, the idea being that we just won't link them from the website anymore.

PR #156, opened 2018-05-04
Add admonition re: FPR server to 1.6 docs

sallain:
Note that a previous PR pulled in the changes from 1.7. If the admonition looks good, I'll apply it verbatim to previous versions of the docs.

Related to #86.

Also fixes a small formatting issue.


sromkey (approved):
Looks good!

PR #164, opened 2018-05-17
Use literalinline for bash snippets

scollazo:
This is a proof of concept in order to use the archivemtica-docs as source of truth for installs.

Instead of using `..code::bash` blocks, this pr uses `..literalinclude::` in order to load the bash snippets from external files.

The current naming scheme for the snippets was designed having in mind the following approach for obtaining the list of snippets needed to run an install:

<pre>
find . -iname 'install_centos' | sort
./admin-manual/installation-setup/installation/scripts/10_install_centos-
yum_upgrade.sh
./admin-manual/installation-setup/installation/scripts/20_install_centos-selinux
.sh
./admin-manual/installation-setup/installation/scripts/30_install_centos-
epel_repo.sh
./admin-manual/installation-setup/installation/scripts/40_install_centos-
elasticsearch_repo.sh
./admin-manual/installation-setup/installation/scripts/50_install_centos-
archivematica_repo.sh
</pre>

A single script can be obtained with `find . -iname '*install_centos*' -print0  | sort -z | xargs -r0 cat`

More info about ..literalinclude::  can be found [here](http://www.sphinx-doc.org/en/stable/markup/code.html#includes)

### DIFF
--- a/admin-manual/installation-setup/installation/install-centos.rst
+++ b/admin-manual/installation-setup/installation/install-centos.rst
@@ -20,33 +20,23 @@ Installation instructions
 
    Update your system
 
-   .. code:: bash
-
-      sudo yum update
+   .. literalinclude:: scripts/10_install_centos-yum_upgrade.sh
+      :language: bash
 
    If your environment uses SELinux, at a minimum you will need to run the
    following commands. Additional configuration may be required for your local
    setup.
 
-   .. code:: bash
-
-      # Allow Nginx to use ports 81 and 8001
-      sudo semanage port -m -t http_port_t -p tcp 81
-      sudo semanage port -a -t http_port_t -p tcp 8001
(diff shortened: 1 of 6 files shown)


sallain (commented):
I like this a lot. One thing that I would love to see included is more information about the scripts. Since it looks the same as the codeblock method if you view the page as HTML, there's nothing to indicate to a user that the scripts exist. A short descriptor about the scripts plus some basic info on accessing and using them would be great (maybe a separate page?). It's not really required to approve this PR but it would be a helpful addition.

scollazo:
As @sallain says, if we follow this path, adding a note pointing to the scripts, and may be create the full script as part of the Makefile that updates the docs, would be nice

scollazo:
Replaced by #228

PR #180, opened 2018-06-13
Adjusts import metadata page for mix of objects and directories

sromkey:
Previously it was not obvious that you could have a mix of objects
and directories in metadata.csv. Several other small adjustments
to reflect functionality: first column can be named filename
regardless of compound/simple objects, dcterms are allowed, removed
reference to dc metadata exporting to AtoM or ArchivesSpace.


sallain (approved):

kellyannewithane (commented):
Looks good @sromkey - I wonder if we should have an example of a csv file with both simple and compound object metadata?

PR #188, opened 2018-08-09
Documentation for resolving decision points

sallain:
This addition to the documentation details how to use the MCPClient to resolve
hanging decision points, which are caused by users removing a transfer or SIP
from the dashboard before resolving a decision that is awaiting user input.

It also fixes up the links to use inline literals and fixes some code-block
formatting.

@mamedin, can you double-check the code snippets and provide corrections if
needed? I didn't have an Archivematica instance to work with to replicate
the steps so I took an educated guess - I'm sure there are errors.

Connected to archivematica/Issues#67


admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,188 @@ edit a configuration file from the command line.
...
+
+It is possible to `remove transfers or SIPs`_ from the Transfer or Ingest tab
+without resolving a decision, which results in the red circle remaining even if
+there is no decision awaiting user input on the tab. While this isn't a critical
+issue, it is annoying. To get rid of the red circle, you need to use a command
+line tool to resolve the user input first and then either complete or reject the
+transfer/SIP. In the Archivematica source code, there is a python module
+containing an RPC client that can give you information about the transfers and
+SIPs that you have in your dashboard. It also allows you to add user inputs, as
+an alternative to using the dashboard.
```
sevein, line 394: Reminder: add a warning somewhere here mentioning that `mcp-rcp-cli` may be replaced in future releases.
Related: https://github.com/archivematica/Issues/issues/68.

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,188 @@ edit a configuration file from the command line.
...
+
+   CentOS:
+
+      .. code-block:: console
+
+         sudo su - archivematica -s /bin/bash
+         source  /usr/share/archivematica/virtualenvs/archivematica-mcp-client/activate
+         export $(grep -v ^# /etc/sysconfig/archivematica-mcp-client)
+         cd /opt/archivematica/archivematica-devtools/bin/
+         ./am mcp-rpc-cli
```
sevein: @sallain @mamedin is this snippet going to work for both Ansible and package-based deploys?

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,188 @@ edit a configuration file from the command line.
...
+.. note::
+
+  If you are running Archivematica on CentOS/Redhat, you must first clone the
+  `archivematica-devtools`_ repo to use this and other tools.
+
+1. Navigate to the Archivematica MCP Client directory.
+
+   Ubuntu:
+
+      .. code-block:: console
```
mamedin: To be consistent with the other commands, I think it is better to use these commands:

Assuming archivematica-devtools was deployed on default directory: /opt/archivematica/archivematica-devtools.

AM 1.7: it works on CentOS and Ubuntu (rpm or ansible deployments):
```   AM 1.7:

          set -a -e -x
          source /etc/default/archivematica-mcp-client || \
              source /etc/sysconfig/archivematica-mcp-client \
                  || (echo 'Environment file not found'; exit 1)
          cd /opt/archivematica/archivematica-devtools/bin/
          ./am mcp-rpc-cli
      ";
```
AM 1.6: (CentOS and Ubuntu)
```
          set -a -e -x
          source /usr/share/python/archivematica-mcp-client/bin/activate
          cd /opt/archivematica/archivematica-devtools/bin/
          ./am mcp-rpc-cli
          deactivate
      ";
```

On Vagrant deployments, archivematica-devtools is installed at `/vagrant/src/archivematica-devtools`so the above commands are valid using:

```
cd /vagrant/src/archivematica-devtools/bin/
```
instead of:
```
cd /opt/archivematica/archivematica-devtools/bin/
```

admin-manual/maintenance/maintenance.rst
mamedin: Yes it works, we are using /etc/sysconfig or /etc/default in all deploys. I proposed on last comment another command that does not use the `sysconfig/default' file for AM 1.6 deployments.

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,188 @@ edit a configuration file from the command line.
...
+issue, it is annoying. To get rid of the red circle, you need to use a command
+line tool to resolve the user input first and then either complete or reject the
+transfer/SIP. In the Archivematica source code, there is a python module
+containing an RPC client that can give you information about the transfers and
+SIPs that you have in your dashboard. It also allows you to add user inputs, as
+an alternative to using the dashboard.
+
+.. note::
+
+  If you are running Archivematica on CentOS/Redhat, you must first clone the
```
mamedin: The prerequisite of having the devtools installed is for all kinds of deployments, not only for CentOS/Red Hat. 
I think it should be clear that this is a requirement for all deployments.

admin-manual/maintenance/maintenance.rst
sallain: Done.

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,194 @@ edit a configuration file from the command line.
...
+
+.. IMPORTANT::
+
+   The tool that is currently being used for this workflow, mcp-rpc-cli, may be
+   deprecated or replaced in future releases. These instructions are accurate as
+   of Archivematica 1.7.2.
+
+.. note::
+
+  If you are running Archivematica on CentOS/Redhat, you must first clone the
```
mamedin: I think it is a general requisite, not only for CentOS

mamedin (changes requested):

admin-manual/maintenance/maintenance.rst
sallain: Where should the devtools repo be cloned to?

sallain:
@mamedin @sevein updated with your changes. I didn't add the 1.6 instructions - once this is merged, I'll cherrypick it back to the 1.6 docs and amend the instructions as needed.

Take a look and let me know if anything else needs fixing.

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,212 @@ edit a configuration file from the command line.
...
 
 4. Save the fileservices
 
 5. Restart Apache
 
-.. code:: bash
+   .. code:: bash
 
-    sudo apache2ctl restart
+      sudo apache2ctl restart
```
sevein, line 361: @mamedin is this still accurate? AFAIK we don't use Apache anymore?

sevein (approved):
LGTM

admin-manual/maintenance/maintenance.rst
sevein: (this could be addressed separately of course)

sevein: Also we would need to warn the user about this change, it should not be done in production. And we'd need to reload Gunicorn which can be done using a signal. I guess that I'll file an issue!

sallain: We can address this as part of work on archivematica/Issues#81

admin-manual/maintenance/maintenance.rst
```diff
@@ -343,39 +342,212 @@ edit a configuration file from the command line.
 2. With your preferred text editor, open the dashboard settings file (vim shown
    in example; Note that you must use sudo to edit this file):
 
-.. code:: bash
+   .. code:: bash
 
-    sudo vim /usr/share/archivematica/dashboard/settings/common.py
+      sudo vim /usr/share/archivematica/dashboard/settings/common.py
```
mamedin, line 347: This file doesn't exist.

admin-manual/maintenance/maintenance.rst
mamedin: yes, this section is totally outdated. I think it is better to fix in a separate issue/PR

mamedin (approved):
LGTM

PR #196, opened 2018-09-13
Add integration page for ArchivesSpace integration

sallain:
This change adds an integration page for the ArchivesSpace integration, similar
to the integration page for the AtoM integration.

Connected to archivematica/issues#149


kellyannewithane (approved):
Looks good, thanks @sallain !

PR #202, opened 2018-09-25
Update CentOS packaging details to 1.8 

joel-simpson:
Update Installation instructions (update url for packages from 1.7 to 1.8)
Update the Upgrade instructions (update the command to go from 1.7 to 1.8)

connected to https://github.com/archivematica/Issues/issues/203
(This PR only partially resolves that issue as it only covers CentOS. Another PR will be required to update details for Ubuntu.)

### DIFF
--- a/admin-manual/installation-setup/installation/install-centos.rst
+++ b/admin-manual/installation-setup/installation/install-centos.rst
@@ -69,24 +69,24 @@ Installation instructions
    * Archivematica:
 
    Use these commands to install the repositories:
-
+      
    .. code:: bash
-
+      
       sudo -u root bash -c 'cat << EOF > /etc/yum.repos.d/archivematica.repo
       [archivematica]
       name=archivematica
-      baseurl=https://packages.archivematica.org/1.7.x/centos
+      baseurl=https://packages.archivematica.org/1.8.x/centos
       gpgcheck=1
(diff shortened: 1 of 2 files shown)


admin-manual/installation-setup/installation/install-centos.rst
```diff
@@ -69,24 +69,24 @@ Installation instructions
    * Archivematica:
 
    Use these commands to install the repositories:
-
+
```
mamedin: I think it is not so important as python code, but an empty line has been replaced by a line with a tab (or several blank spaces). Same for lines 74 and 83

mamedin (changes requested):
It is about CentOS install and upgrade, so PR title and commit messages should mention it is only for CentOS.

admin-manual/installation-setup/installation/install-centos.rst
joel-simpson: I updated to fix these so they have empty lines as before.

mamedin (changes requested):
Sorry, but lines 74 and 83 still have blank spaces :(

joel-simpson:
sorry @mamedin I didn't check properly the first time. I think I've fixed it properly this time.

mamedin (approved):
LGTM

sallain (approved):

sallain:
Merged!

PR #204, opened 2018-09-26
Move old FPR document to main docs repo and update

sallain:
This change adds administrator-level FPR documentation to the main AM docs
repository. Prior to 1.7, we maintained the FPR docs as a branch within
artefactual/archivematica-fpr-admin. I've also substantially updated both the
contents and the formatting of this documentation, though I'm sure there is
still clean-up left to do.

Connected to archivematica/Issues#205. This change will check off the first
to-do item in that issue.


sallain:
Cancel this, I'm rethinking some things. Somehow I got it into my head that @sromkey's changes to the preservation planning docs had missed some things from the FPR docs, but now I'm thinking that I misread stuff and essentially duplicated Sarah's work. Will disambiguate and create a new PR.

Issue #33, opened 2017-10-06
update docs on how to install without elasticsearch


kellyannewithane:
Problem: Need to document in admin manual and other relevant places how to install Archivematica without ElasticSearch

kellyannewithane:
1.8 branch with updated documentation is ready for review

sallain:
I'm reasonably confident that this was completed and that the changes are still present in the current version of the 1.8 documentation, even though we've substantially changed the installation pages. The only page that does not contain Elasticsearch instructions is [Automated installation using Ansible](https://www.archivematica.org/en/docs/archivematica-1.8/admin-manual/installation-setup/installation/install-ansible/) - should the optional Elasticsearch information be included on this page?

@sevein pinging you, though I'm not sure if you're the right person to ask.

sevein:
@sallain there are instructions here: https://github.com/artefactual-labs/ansible-archivematica-src/#disable-elasticsearch-use. Perhaps the docs should point to that README because it contains other useful information.

sallain:
Thanks @sevein! I'll update the docs.

sallain:
Documented and added to release notes

PR #207, opened 2018-10-09
Add Elasticsearch info to ubuntu-ansible docs

sallain:
This change adds a note to the ubuntu-ansible installation documentation
instructing users how to install Archivematica without Elasticsearch when using
the ansible role.

Connected to #33.


sevein (approved):
:tada:

PR #212, opened 2018-10-19
Add Binder integration information

sallain:
This change covers a lot of things:

* Creates an integrations landing page for Binder
* Updates the dashboard configuration pages (admin and user manuals)
* Adds instructions for uploading a DIP to Binder
* Adds instructions for using the Access system ID box for both AtoM and Binder
* Adds screenshots of various things

Plus my usual formatting updates to bring the pages in line with the style guide.

Connected to archivematica/Issues#23


sallain:
@jraddaoui I added you as a reviewer because there is a lot of configuration information about Binder that is currently missing - details about how to make Binder, Archivematica, and TMS talk to each other. I don't think we need to address that for 1.8, as it's very unlikely that someone is going to be setting up Binder without our help, but if there's anything that's really important that's missing, let me know!

ablwr (approved):

PR #217, opened 2018-10-25
Adds an admonition re: using the correct identifiers for Binder

sallain:
Connected to archivematica/Issues#23


kellyannewithane (approved):
beautiful!

PR #218, opened 2018-10-25
Update translations page

sallain:
This change is a bit of a tidy-up to the translations page in the docs, as well
as the Language section of the dashboard config pages.

Connected to archivematica/Issues#231


sevein:
> Connected to #231

Remember to use the full link otherwise it won't work.

sallain:
@sevein *facepalm*

currmie (commented):
Just one minor recommendation--
I think there are 4 instances of "click save" in the file (if you do a text search). Perhaps change those instances to "click **Save Translation**" to reflect exact wording from the Transifex UI.

sallain:
@currmie thanks! I have incorporated that suggestion.

Issue #114, opened 2018-03-28
Problem: Forensic image unpacking document vague


ablwr:
Version of the documentation:
1.7

Page (and section, if applicable) where the issue occurs:
https://www.archivematica.org/en/docs/archivematica-1.7/user-manual/transfer/forensic/

Description of the issue:
In working with Eddy Colloton, we've determined that while ISO and RAW disk images are well-suited for ingest and extraction within Archivematica (post ingest), E01 files are not so well-suited. In order for E01 files to be ingested, the FPR would have to be updated and given a more complex script than the one provided (to simplify -- an earlier script by The Sleuth Kit would have to be run to gather information, that information would then have to be reconstructed as temporary variables that would then pass into the script that AM currently runs for unpacking). The E01 files do not provide enough information for the current Sleuth Kit script to run on its own. The documentation, too, is ambiguous in this regard, which leads to the false implicit understanding that E01 disc images can, too, be unpacked after ingest.

To create a user story:
As a preservationist, I would like to be aware of which types of disc images can be sufficiently unpacked within Archivematica.

I defer to @sallain advice on how to best proceed here (if anything), and am happy to send an update indicating that RAW and ISO are supported but E01 are not.

ablwr:
Also, Eddy is using BitCurator to generate these, which is noted in the documentation. BitCurator exports RAW and E01.

sromkey:
I don't disagree about the docs, but I thought we had tackled this problem for UCLA. I'm pretty sure at least some E01 images can be handled by the command in the FPR?

ablwr:
(Based on me lurking on a convo on twitter between Eddy and Shira, my understanding is..) UCLA doesn't try to extract them after, and Eddy is trying to. So they can go in and through just fine, but they can't be used in the extract microservice.

sromkey:
Got it, thanks for clarifying!

ablwr:
Yeah... trying to figure out how best to articulate the issue succinctly and without getting into the mess of listing out file formats and all their nuances. I think it could be as simple as "For BitCurator integration, only exported RAW files can use the Extract microservice."

sallain:
@ablwr I think you can go ahead and make a PR for that change, if you want!

@sromkey Do you think there's anything else to do on this? Just clearing up my issue assignments/mentions.

sallain:
@ablwr do you think that this issue has been resolved? If not, happy to pick it back up!

ablwr:
Let's close this!

PR #236, opened 2018-12-10
Add extraction limitation for forensic images #114

ablwr:
I had attempted to fix this many months ago https://github.com/artefactual/archivematica-docs/pull/151/files but it didn't make the cherry-picking cut, I think. So I am bringing it back with the original sentence constructed in #114 and it will also resolve that issue.


sallain (approved):
LGTM! Funny that it didn't make it into the 1.8 docs :thinking:

PR #237, opened 2018-12-11
Include MediaConch policy creation instructions

ablwr:
This takes the pertinent information from [the wiki](https://wiki.archivematica.org/MediaConch_workflow) on how to use MediaConch and updates with an accurate minimal policy example. It skips the long-winded one. This will resolve https://github.com/archivematica/Issues/issues/374 and the wiki page can be removed and/or redirected to the docs. 

I think there's more I can do to improve these docs!!! I am going to work more on that as I tackle training and bug-fixing obligations associated with MediaConch. But for the basic goal of the issue, this suffices.

### DIFF
--- a/user-manual/preservation/preservation-planning.rst
+++ b/user-manual/preservation/preservation-planning.rst
@@ -829,9 +829,96 @@ Validation tools
 ^^^^^^^^^^^^^^^^
 
 Archivematica includes two validation tools: `JHOVE`_ and `MediaConch`_.
+
+
+MediaConch
+^^^^^^^^^^
+
 MediaConch was introduced in Archivematica 1.7 as a validation tool specifically
-for Matroska (MKV) files, opening up the possibility to include more specialized
-tools in future releases.
+for files utilizing the Matroska (MKV) container format, the FFV1 video codec
+format, and the LPCM audio codec format. This is the file format used and 
+recommended by Archivematica when performing normalization on audiovisual files.
+For more information about normalization rules, see the :ref:`Normalization
+<normalization>` section below.
+
+In addition to validating files against their format specification, MediaConch 
+can also be used to validate file formats against an internal policy created by 
+any institution. 
+
+Users can use `MediaConchOnline`_ or a local installation of MediaConch to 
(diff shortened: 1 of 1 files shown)


user-manual/preservation/preservation-planning.rst
```diff
@@ -829,9 +829,66 @@ Validation tools
...
+`POLICY = """` and `""".strip()` in
+the Command section. As an example, here's a very simple policy to check that a 
+file is an mp4:
+
+.. code-block:: none
+
+   <policy type="and" name="Is Mp4?" license="CC-BY-SA-4.0+">
+     <description>New Is Mp4</description>
+     <rule name="Is Mp4?" value="Format" tracktype="General" occurrence="*" operator="=">mp4</rule>
+   </policy>
```
sallain, line 898: I wonder if in this example it would be helpful to include the surrounding text as well? My concern is that someone might not read thoroughly and just include the policy text, erasing the code around it.

user-manual/preservation/preservation-planning.rst
```diff
@@ -829,9 +829,66 @@ Validation tools
...
+
+3. Create or replace a rule (in the Rules section under Validation) with the 
+purpose "Validation against a Policy" for the format you want to check, using 
+the command you just created/edited. 
+
+4. Now, when processing a new transfer, there should be a microservice called 
+"Policy checks on originals." Clicking on the gear icon associated with 
+that microservice will give you details related to your MediaConch policy. Like 
+with other microservices in Archivematica, its background will be green when all 
+policies pass, and turn pink if one or more policies fail.
```
sallain: I think adding some text about the policy check against derivative options would be good too!

sallain (approved):
Looks good! I made a couple of suggestions, but you can adopt them or not as you see fit/have time.

user-manual/preservation/preservation-planning.rst
ablwr: OK I have an update coming momentarily after I check the build!

ablwr:
@sallain Thank you! I force pushed a new update to keep it to one commit and took your suggestions!

PR #240, opened 2019-01-04
Add transfer wont start help

ablwr:
Some tips (mostly written by you @sallain)!

### DIFF
--- a/admin-manual/maintenance/maintenance.rst
+++ b/admin-manual/maintenance/maintenance.rst
@@ -544,6 +544,51 @@ an alternative to using the dashboard.
 6. To update the list and continue, enter `u` when prompted to enter a value.
    Repeat the steps above until there are no more choices left to make.
 
+
+.. _transfer-wont-start:
+
+Transfer won't start
+^^^^^^^^^^^^^^^^^^^^
+    "I try to create a new transfer, but nothing happens. What can I do?"
+
+Sometimes a user may attempt to start a transfer and it will never seem to 
+initiate the Archivematica processes. There are a few issues to look out for 
+and investigate if this happens.
+
+1. File permissions
+
+  First, the issue may be related to file permissions in the transfer source 
+  directory. Check the permissions in the directory and on the files to ensure 
+  that all files can be read by Archivematica.
+
+2. System timeouts
+
+  If it is a large transfer, it may just be taking a long time to copy the files
+  and initially load them into the system, and the user can wait a bit longer 
+  and see if the processes begin after a bit of time. It is also possible that 
+  it is taking a long time because some of the system timeouts are being 
+  exceeded and the transfer has failed. This can be verified by checking the 
+  Storage Service logs and by checking where the transfer exists on the 
+  filesystem.
+
+  For inadequate timeouts, check the Storage Service configuration and adjust 
+  if necessary.
+
+3. Communication between Dashboard and Gearman
+
+  If the transfer has successfully moved to the shared Directory (i.e. it can be
+  found in ``sharedDirectory/watchedDirectories/activeTransfer/`` folders), but 
+  is still not showing up in the dashboard, there could have been a problem with 
+  the communication between the dashboard and Gearman. Restarting all of the 
+  services can resolve this problem and the transfer will appear.
+
+  Restart services in the follow order: ``gearmand``, 
+  ``archivematica-mcp-server``, ``archivematica-mcp-client``, 
+  and ``archivematica-dashboard``.
+
+  Note that on some installations, ``gearmand`` may be called 
+  ``gearman-job-server``.
+
 :ref:`Back to the top <maintenance>`
 
 .. _`pyes`: https://github.com/aparo/pyes/


ablwr:
I don't understand why this fails. :( Locally, builds fine and looks fine!

sallain:
```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
The command "sphinx-build -n -W -b html -d _build/doctrees . _build/html" exited with 2.
```

Probably need @sevein to take a look at this???

ablwr:
Oh, if I run `make clean html` instead of just `make html` I do get some similar errors. But you just sent in a PR that worked totally fine, which is why I'm confused now!

sevein (commented):
@ablwr, you forgot to truncate the text after 80 columns.

ablwr:
Oh man I would love an auto-linter for the docs. Ok, sorry about that, will update!

sevein (commented):
@ablwr, I've just noticed how the other sections under this FAQ are titled as a question or they're started with the question quoted in the body. Should we do the same for consistency?

I also think that you need to review your indentation because rst is very picky rendering that stuff. Two spaces are used under "File permissions", a mix of 2 and 3 spaces under "System timeouts", etc...

---

image

admin-manual/maintenance/maintenance.rst
```diff
@@ -544,6 +544,58 @@ an alternative to using the dashboard.
...
+  Storage Service logs can be found in the 
+  `/var/log/archivematica/storage-service/` directory.
+
+  For inadequate timeouts, check the Storage Service configuration and adjust 
+  if necessary.
+
+3. Communication between Dashboard and Gearman
+
+  If the transfer has successfully moved to the shared Directory (i.e. it can be
+   found in sharedDirectory/watchedDirectories/activeTransfer/ folders), but it
```
sevein: `sharedDirectory/watchedDirectories/activeTransfer/` could be an inline literal or you could also use that other form that I can't remember how it's called, e.g. `:file:`sharedDirectory/dadada...` (interpreted text?).

admin-manual/maintenance/maintenance.rst
```diff
@@ -544,6 +544,58 @@ an alternative to using the dashboard.
...
+  If it is a large transfer, it may just be taking a long time to copy the files
+   and initially load them into the system, and the user can wait a bit longer 
+   and see if the processes begin after a bit of time.  It is also possible that 
+   it is taking a long time because some of the system timeouts are being 
+   exceeded and the transfer has failed. This can be verified by checking the 
+   Storage Service logs and by checking where the transfer exists on the 
+   filesystem.
+
+  Storage Service logs can be found in the 
+  `/var/log/archivematica/storage-service/` directory.
```
sevein: I don't know if it really makes sense to have interpreted text without a role (e.g. `file` role). You could either use:

    :file:`/var/log/archivematica/storage-service/`

Or you could use an inline literal:

    ``/var/log/archivematica/storage-service/``

See [inline markup](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#inline-markup) for more details.

admin-manual/maintenance/maintenance.rst
```diff
@@ -544,6 +544,58 @@ an alternative to using the dashboard.
...
+
+  First, the issue may be related to file permissions in the transfer source 
+  directory. Check the permissions in the directory and on the files to ensure 
+  that all files can be read by Archivematica.
+
+2. System timeouts
+
+  If it is a large transfer, it may just be taking a long time to copy the files
+   and initially load them into the system, and the user can wait a bit longer 
+   and see if the processes begin after a bit of time.  It is also possible that
```
sevein: Double sentence spacing not needed.

admin-manual/maintenance/maintenance.rst
sevein: To be fair the logging config depends on the distribution. There is a comprehensive description here that you could use instead or in addition to what you've already said: https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/install/README.md#logging-configuration.

admin-manual/maintenance/maintenance.rst
```diff
@@ -544,6 +544,58 @@ an alternative to using the dashboard.
...
+   services can resolve this problem and the transfer will appear.
+
+  Restart services in the follow order:
+
+    .. code-block:: bash
+
+      service gearmand restart
+      service archivematica-mcp-server restart
+      service archivematica-mcp-client restart
+      service archivematica-dashboard restart
```
sevein: Is this command compatible with all the distros supported? It's totally possible but I'm not entirely sure.

sevein (commented):
Some more comments :sweat_smile:

ablwr:
Thank you @sevein!! I'll clean this up.

admin-manual/maintenance/maintenance.rst
sallain: Inline literals are what we usually use!

admin-manual/maintenance/maintenance.rst
sallain: +1 for inline literal in this situation.

admin-manual/maintenance/maintenance.rst
```diff
@@ -544,6 +544,58 @@ an alternative to using the dashboard.
...
+
+  Note that on some installations, `gearmand` may be called 
+  `gearman-job-server`.
+
```
sallain, line 591: Change these to inline literals as well (surround with double-backticks).

sallain (commented):
A few small changes, and I think it's worth looking into sevein's comment re: distros, but otherwise looks good!

ablwr:
I made this a bit more minimal in order to be a bit more system-agnostic, which is probably good overall in case systems in the future don't act the same as they do now, and there will be less things to track down and change. Sorry about the backticks issue! Hard to undo 7 years of github-flavored markdown burned into my mind! :dizzy_face:

sevein (approved):

PR #241, opened 2019-01-10
Update metadata.csv transfer page

ablwr:
This tidied up the metadata.csv transfer page a bit and also moves the samples from in-line to attachments.


user-manual/transfer/import-metadata.rst
```diff
@@ -54,43 +62,46 @@ MDTYPE="OTHER". A sample of the METS output is available below.
...
-     directory: e.g. "objects/BrocktonOval.jp2" , or "objects/Jan021964".
+   * The first column in the ``metadata.csv`` file must be a "filename" column.
+     This column should list the filepath and filename (e.g. 
+     "objects/BrocktonOval.jp2") or directory name of each object or
+     directory (e.g. "objects/Jan021964").
 
-   * If you have directory level metadata fill out the fields on the same line
-     as the directory (ie objects/)
+   * If you have directory level metadata, fill out the fields on the same line
+     as the directory (e.g. objects/).
```
sallain, line 106: Maybe there should there be an admonition somewhere about the fact that you need to include ``objects/`` in the path even if the materials aren't in an objects subfolder? It's a legacy artifact from the days when all digital objects in the transfer had to be nested in such a way.

sallain (approved):
One sort of general comment, but otherwise looks good to me!

user-manual/transfer/import-metadata.rst
ablwr: I agree with this!!! I thought about that and forgot to add it in.

PR #243, opened 2019-01-11
Restore formatting patch

ablwr:
Picky Sphinx!


sromkey (approved):
This should do it I think!

PR #245, opened 2019-01-17
Add more information about name cleanup

sallain:
This is an attempt to explain why and how Archivematica changes file
and directory names during Transfer and Ingest. Feedback welcome!

Connected to archivematica/Issues#387

### DIFF
--- a/user-manual/transfer/_csv/file-name-edits.csv
+++ b/user-manual/transfer/_csv/file-name-edits.csv
@@ -1,0 +1,6 @@
+Original name,Edited name,METS output
+i & tem,i___tem,"Original name=""%transferDirectory%objects/i &amp; tem.png""; cleaned up name=""%transferDirectory%objects/i___tem.png"""
+Česká republika.png,Ceska_republika.png,"Original name=""%transferDirectory%objects/Česká republika.png""; cleaned up name=""%transferDirectory%objects/Ceska_republika.png"""
+Éireann,Eireann,"Original name=""%transferDirectory%objects/Éireann.png""; cleaned up name=""%transferDirectory%objects/Eireann.png"""
+España,Espana,"Original name=""%transferDirectory%objects/España.png""; cleaned up name=""%transferDirectory%objects/Espana.png"""
+Росси́я,Rossiia,"Original name=""%transferDirectory%objects/Росси́я.png""; cleaned up name=""%transferDirectory%objects/Rossiia.png"""
--- a/user-manual/transfer/clean-up-names.rst
+++ b/user-manual/transfer/clean-up-names.rst
@@ -9,16 +9,17 @@ Transfer tab and once on the Ingest tab. The purpose of this microservice is to
 ensure that Archivematica's tools and processes do not fail because of
 characters that appear in file or directory names.
 
-Archivematica groups a wide variety of tools together to create preservation
+Archivematica groups a wide variety of tools, such as :ref:` preservation
+planning tools <preservation-planning-tools>`, together to create preservation
 workflows. Some of the tools that Archivematica uses have a narrowly-defined
 scope in terms of what the tool considers to be a valid character for a file or
 directory name. Encountering a character that the tool considers to be invalid
 can cause the tool to fail, which can halt processing and interfere with normal
 operation of Archivematica.
 
-To prevent these kinds of tool failures from happening, Archivematica implements
-a script that changes file and directory names to conform to requirements of the
-most restrictive tools. Valid characters are defined by `specific code`_ in
+To prevent these kinds of failures from happening, Archivematica implements
+a script that changes file and directory names to conform to the requirements of
+more restrictive tools. Valid characters are defined by `specific code`_ used by
 Archivematica:
 
 ``-_.()abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789``
--- a/user-manual/transfer/clean-up-names.rst
+++ b/user-manual/transfer/clean-up-names.rst
@@ -1,0 +1,91 @@
+.. _clean-up-names:
+
+==============
+Clean up names
+==============
+
+The clean up names microservice runs twice during processing - once on the
+Transfer tab and once on the Ingest tab. The purpose of this microservice is to
+ensure that Archivematica's tools and processes do not fail because of
+characters that appear in file or directory names.
+
+Archivematica groups a wide variety of tools together to create preservation
+workflows. Some of the tools that Archivematica uses have a narrowly-defined
+scope in terms of what the tool considers to be a valid character for a file or
+directory name. Encountering a character that the tool considers to be invalid
+can cause the tool to fail, which can halt processing and interfere with normal
+operation of Archivematica.
+
+To prevent these kinds of tool failures from happening, Archivematica implements
+a script that changes file and directory names to conform to requirements of the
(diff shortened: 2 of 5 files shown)


sallain:
Travis failed but I really don't know why! I've tried googling but no luck - it seems to have something to do with the new CSV I added, but it's the same as other places we have CSVs (see admin-manual/installation, for ex.)

user-manual/transfer/clean-up-names.rst
```diff
@@ -0,0 +1,93 @@
+.. _clean-up-names:
+
+==============
+Clean up names
+==============
+
+The clean up names microservice runs twice during processing - once on the
+Transfer tab and once on the Ingest tab. The purpose of this microservice is to
```
ablwr, line 8: Sort of a general style guide question -- do we usually say "Transfer tab" and "Ingest tab" or "during Transfer" or some other archivally-suited way? I think how you are saying it is fine, just wondering about alternatives I've used in the past and what we generally should aim for when describing digipres software!

ablwr (approved):
Well, I think this looks great, even if Travis doesn't. When it was failing for me, one time it was a stray space that I couldn't see that was beyond the 70-char mark, and it builds fine locally too.

user-manual/transfer/_csv/file-name-edits.csv
```diff
@@ -0,0 +1,8 @@
+Original name,Edited name,METS output
+@file,_file,"Original name=""%transferDirectory%objects/@file.png""; cleaned up name=""%transferDirectory%objects/_file.png"""
+file^,file_,"Original name=""%transferDirectory%objects/file^.png""; cleaned up name=""%transferDirectory%objects/file_.png"""
```
replaceafill: @sallain I think these two `file_` strings in this line are being interpreted as [hyperlink references](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#inline-markup). Try escaping them as `file\_`

sallain: I tried that and it still didn't work! So I've just removed the offending lines.

user-manual/transfer/clean-up-names.rst
sallain: I usually say "on the Transfer tab" or "on the Ingest tab" for clarity, especially when it's a job/task that appears on both. But I think other alternatives are fine, as long as it's obvious where we're all supposed to be looking!

user-manual/transfer/clean-up-names.rst
```diff
@@ -0,0 +1,91 @@
...
+ensure that Archivematica's tools and processes do not fail because of
+characters that appear in file or directory names.
+
+Archivematica groups a wide variety of tools together to create preservation
```
ross-spencer: I tried to see if we say this anywhere else in the docs (it seems like something we might say!) but I could only find this: Should we link to it? https://www.archivematica.org/en/docs/archivematica-1.8/getting-started/overview/microservices/ 

Ooh! Or maybe this one? https://www.archivematica.org/en/docs/archivematica-1.8/user-manual/preservation/preservation-planning/#tools

user-manual/transfer/clean-up-names.rst
```diff
@@ -0,0 +1,91 @@
...
+workflows. Some of the tools that Archivematica uses have a narrowly-defined
+scope in terms of what the tool considers to be a valid character for a file or
+directory name. Encountering a character that the tool considers to be invalid
+can cause the tool to fail, which can halt processing and interfere with normal
+operation of Archivematica.
+
+To prevent these kinds of tool failures from happening, Archivematica implements
```
ross-spencer:
```suggestion
To prevent these kinds of failures from happening, Archivematica implements
```
At this point I feel like the context is set, we don't need to use tool again for the moment. I suggested we take it out, but do you think reads okay?

user-manual/transfer/clean-up-names.rst
```diff
@@ -0,0 +1,91 @@
...
+a script that changes file and directory names to conform to requirements of the
+most restrictive tools. Valid characters are defined by `specific code`_ in
```
ross-spencer:
```suggestion
more restrictive ones. Valid characters are defined by `specific code`_ in
```
If we use the change above, then we'd need a pronoun here. It might work?

user-manual/transfer/clean-up-names.rst
```diff
@@ -0,0 +1,91 @@
...
+``-_.()abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789``
+
+The script tries to replace any character that is not included in the above list
+with its nearest equivalent - for example, the character ``Č`` is replaced by
+``C``. Where the script cannot find a close equivalent, the character is
+replaced by an underscore. The change is documented in the METS file.
+
+.. csv-table::
+   :file: _csv/file-name-edits.csv
+   :header-rows: 1
```
ross-spencer, line 34: This is a very cool way to include a table!

user-manual/transfer/clean-up-names.rst
ross-spencer: > Valid characters are defined by specific code in Archivematica:

I see what you mean. There might be a way we can be more precise for the reader here, not just to help us maintain the sensitivity with which you've approached these docs, but to open up a greater possibility of change as well.

If we say:

`Valid characters are defined by specific code used by Archivematica:`

The context is that outside of these docs we can better clarify that the limited set below is not entirely a design choice by our team. The fact that these characters is *such* as narrow band comes from the standard Python library we use: https://docs.python.org/3/library/string.html#string.ascii_lowercase 

We could add the detail, but I'm only advocating for `in` to be replaced with `used by`.

ross-spencer:
Hi @sallain just a few comments in case they're useful. Otherwise I really like the way you've approached writing this. It'll be a really useful resource once merged.

user-manual/transfer/clean-up-names.rst
sallain: I think "more restrictive tools" would read better?

user-manual/transfer/clean-up-names.rst
sallain: @currmie taught us this! :tada:

user-manual/transfer/clean-up-names.rst
sallain: I think this is a good suggestion! Can you suggest the change? I want to make use of this fun feature :smile_cat:

ross-spencer: Hahaha! I feel like I'm being a party-pooper, but because my comment and first suggestion are restricted to the one line it won't let me! Sorry :smiley:

sallain: Hehe that's okay. Thanks.

user-manual/transfer/clean-up-names.rst
sallain: Added the second link!

user-manual/transfer/clean-up-names.rst
sallain: Added manually, even though I want to use the suggestion feature so bad.

user-manual/transfer/clean-up-names.rst
sallain: Fixed up manually.

sallain: Fixed!

PR #251, opened 2019-02-05
Change right-hand column mentions to left-hand

peterVG:
fixes issue #484


sallain (approved):
LGTM! Can you make the change in the 1.9 docs too?

PR #252, opened 2019-02-06
Update AtoM integrations setup page

sallain:
This change adds a note that the AtoM REST API must be enabled from the AtoM
plugins menu during setup. This change also tidies up the integration page and
adds a few more links.

Change will be cherry-picked to 1.9 docs once approved.

Connected to archivematica/Issues#487


ablwr (approved):

ablwr:
Woops, sorry I took so long on this!

PR #257, opened 2019-02-19
Update repos for 1.9

scollazo:
This PR updates the repositories used for 1.9.x ,and also bring the inline snippets for ubuntu installs


scollazo:
Connects to https://github.com/archivematica/Issues/issues/525

sallain (approved):
This looks great @scollazo! Thank you so much for putting in the inline scripts!!!

PR #276, opened 2019-04-15
Dev/issue 602 remove sip creation from ingest

peterVG:
SIP arrangement and creation has been removed from the Ingest tab in release 1.10. All SIP creation from backlog will now occur on the Appraisal tab. This requires removing mentions of SIP arrangement on the Ingest User Manual page as well as updates of screencaps that no longer show the SIP arrangement pane.

Connected to https://github.com/archivematica/Issues/issues/602


sallain (approved):
LGTM!

PR #280, opened 2019-04-18
Fixes links to installation docs

sallain:
Some of the links in the table on the installation page were broken,
taking users to the wrong place in the installation docs. I've fixed those
links and also made the tables a little nicer looking by breaking each
section into a separate table. Sphinx's table formatting isn't very advanced
so this was a way to get the headers to look nicer.

Connected to archivematica/issues#661


sromkey (approved):
lgtm :+1:

PR #290, opened 2019-06-05
Add documentation on using JSON for DC metadata

sallain:
Adds a section for using a JSON file rather than a CSV. Also includes an
example JSON file.

Connected to archivematica/Issues#566


ablwr (approved):
Yes!! I recreated and tested using this example!!!

ross-spencer:
@sallain sorry, late to the party on this one. Is it worth adding that elements can still be repeated using this technique, just that the structure has to be changed to that of a JSON-array, e.g. here for [`dc:subject`](https://github.com/artefactual/archivematica-sampledata/blob/master/SampleTransfers/DemoTransferJSON/metadata/metadata.json#L15-L21)?

sallain:
@ross-spencer I didn't mention it explicitly, but the [JSON sample](https://github.com/artefactual/archivematica-docs/pull/290/files#diff-897b3e9fad0626ccf315aec2a0a50a10) does have a repeated element so I hope that it is intuitive! I will be further expanding import-metadata.rst to add the DC Terms info so I'll take it into consideration then as well.

sallain:
As per @ross-spencer's advice I've created a new PR to address repeated elements: https://github.com/artefactual/archivematica-docs/pull/291

PR #291, opened 2019-06-07
Fixed repeating elements

sallain:
Fixes up the metadata.json sample and adds a line about formatting for repeating elements.


cole (approved):
Looks good!

PR #293, opened 2019-06-10
add PID declaration docs

peterVG:


user-manual/transfer/transfer.rst
```diff
@@ -325,6 +326,40 @@ click on the gear icon for the job.
...
    bagging process. You do not need to create checksums manually.
 
+
+.. _transfer-pids:
+
+Transfers with existing persistent identifiers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Archivematica can import persistent identifiers that were created
+outside of Archivematica. Perhaps your workflow includes DOI identifiers or
```
sallain: Would change this to be more neutral, i.e.:

```suggestion
outside of Archivematica, such as DOI identifiers or
```

user-manual/transfer/transfer.rst
```diff
@@ -325,6 +326,40 @@ click on the gear icon for the job.
...
+  :width: 50%
+  :alt: A persistent identifier (PID) is declared as a link to a PREMIS object
+
+Once Ingest completes, the PID will be linked to the premis:objectIdentifier in
+the AIP METS file.
+
+.. image:: images/PID_linked_to_PREMIS.*
+  :align: center
+  :width: 50%
+  :alt: PIDs linked to PREMIS objects
```
sallain, line 361: Any code examples should be included as text, rather than screenshots, so that they are more accessible (i.e. text can be zoomed as big as needed, whereas images get blurry; text can also be read by a screenreader.) This is something I've been slowly changing over time so no need to do it now if you don't have time, though.

For an example, see the METS examples in import-metadata.rst

sallain (approved):
Approved, though I've made a couple of small suggestions that you can implement or not!

sallain:
Connected to archivematica/Issues#133

PR #294, opened 2019-06-24
update arrangement section to reflect Appraisal tab changes

peterVG:
note: that the SIP arrangement feature for 1.10qa was experiencing a bug while writing these docs so the intention is to add one more screencapture to show drag-n-drop under the Add directory section.


peterVG:
Connected to archivematica/Issues#598

sallain (approved):
LGTM! This is going to be such a nice lil upgrade to the Appraisal tab.

PR #298, opened 2019-07-11
Fix CentOS dashboard upgrade instructions

jorikvankemenade:
When running the migration with `--no-input` it fails, changing it to the valid `--noinput` option.

Connects archivematica/issues#699


mamedin (approved):
LGTM

sallain:
Thanks @jorikvankemenade!

PR #301, opened 2019-07-19
Add Gzipped tar to compression options

ross-spencer:
Gzipped tar is the most recent addition to Archivematica's
comression options. Introduced to 1.10 in our work with the Wellcome
Trust. Compression algorithms have also been organized in
alphabetical order to match that seen in the user-interface.


sromkey (approved):
Looks great, thanks for doing this!

PR #306, opened 2019-08-14
Re-write manual normalization page

sallain:
This change clarifies the manual normalization options in Archivematica.
I've substantially re-written the manual normalization page, and also
updated the Transfer page where it mentions manual normalization.

Structure-wise, I've replaced screenshots and also added a csv folder in
order to reference the normalization.csv file, as per the style guide.

Along with a previous change to the transfer page, this PR resolves
archivematica/Issues#669


user-manual/ingest/manual-normalization.rst
```diff
@@ -4,261 +4,357 @@
...
+      ├── hooded-pitcher-plant.tif
+      ├── ornamental-onion.gif
+      ├── ornamental-onion.png
+      ├── ornamental-onion.tif
+      ├── pepperbox-poppy.gif
+      ├── pepperbox-poppy.png
+      └── pepperbox-poppy.tif
+
+   In this example, there are three copies of each file - a GIF, a PNG, and a
+   TIFF.
```
sromkey: It might be clarifying here to give a hint of where this is going - "The PNG files will be used as the original files, the TIF files as preservation derivatives and the GIF files as access derivatives." Or similar.

sromkey: Or alternatively, you could start with an intro that kind of outlines the scenario: "A user has a set of files that uses PNG as the original format, TIF as the preservation derivative and GIF as the access derivative..."

user-manual/ingest/manual-normalization.rst
```diff
@@ -4,261 +4,357 @@
...
+      │       ├── ornamental-onion.tif
+      │       └── pepperbox-poppy.tif
+      ├── hooded-pitcher-plant.png
+      ├── ornamental-onion.png
+      └── pepperbox-poppy.png
+
+   In the example above, each original PNG file has a manually normalized access
+   derivative (in GIF format) and a manually normalized preservation derivative
+   (in TIFF format).
+
```
sromkey, line 114: I guess if you use my suggestion above this statement would no longer be necessary! Either way works I just like to know where things are going as I read :)

user-manual/ingest/manual-normalization.rst
```diff
@@ -4,261 +4,357 @@
...
+#. Return to the dashboard and approve normalization.
 
-If you include manually normalized files in the initial transfer (option 1) or
-during processing (option 2), the procedure for adding PREMIS eventDetail is
-the same.
+#. Once the AIP is created, you will find that the originals and their
+   preservation derivatives are now located in the ``data`` directory of the AIP
+   as usual. Note that the ``manualNormalization`` directory and subfolders were
+   deleted once normalization was complete. If you had access copies, a DIP will
+   be created as well.
```
sromkey, line 238: This has been the best description of this workflow I have ever read! :smile_cat:

sromkey (approved):
I have a suggestion in the comments but I'm also ok with leaving this as is if you disagree. In any case it's made the manual normalization instructions a lot easier to understand so thank you!!

user-manual/ingest/manual-normalization.rst
sallain: Good call!

user-manual/ingest/manual-normalization.rst
sallain: Impressive, since I've never actually done it.

PR #308, opened 2019-08-20
Remove Archivists Toolkit documentation

sallain:
Deletes the Archivist's Toolkit integration page and removes references
to the integration from a handful of other pages and indices.

Connected to archivematica/Issues#174


sromkey (approved):
Looks good! I think that should do it!

PR #312, opened 2019-10-09
Remove ghostscript in rpm updates

scollazo:
Related to https://github.com/archivematica/Issues/issues/915


mamedin:
Hi @scollazo , does this PR assume that ghostscript* packages have been removed from the artefactual CentOS repo ? 

In case the 1.10 rpm repo is not changed we need to run the following additional command because the ImageMagick dependency before upgrading:

```
sudo yum install ghostscript
```

sallain (approved):

scollazo:
@mamedin the package has been removed from the repo

PR #313, opened 2019-10-09
Remove references to elasticsearch upgrade

scollazo:
This only applies to upgrades from 1.8 to 1.9, it doesn't make sense
when updating from 1.9 to 1.10/1.10.1


jraddaoui (commented):
Thanks @scollazo!

I'm not sure how the upgrading docs are formed to upgrade from various versions before (e.g.: 1.8 to 1.10). If we don't indicate the need of checking the previous versions upgrading docs, we may want to add a note in here for those upgrading from 1.8 or lower. Otherwise, it looks good to me.

sallain (approved):

PR #316, opened 2019-10-16
Updated error-handling sections

sarah-mason:
* Created a section called Errors that fail transfers
* Made edits to other common behaviours (including removing transfers that belonged in the previous section)
* All checked section against style guide

This is connected to issue: https://github.com/archivematica/Issues/issues/569

### DIFF
--- a/getting-started/troubleshooting/error-handling.rst
+++ b/getting-started/troubleshooting/error-handling.rst
@@ -226,7 +226,6 @@ a mistake was made entering the permalink:
 
    Warning that permalink was incorrect, allows user to retry upload DIP
 
-
 .. _failure-errors:
 
 Errors that fail transfers
@@ -238,21 +237,6 @@ all microservice errors will lead to a failed transfer. Below is a list of
 common errors that will result in the 'Failed transfer' microservice to run and
 the transfer to be moved to the failed directory.
 
-#. Approve Transfer (zipped and unzipped bags): When running a zipped or
-   unzipped bag transfer, the first microservice is 'Approve transfer.' If the
-   'verify bag, and restructure for complaince' job fails within this
-   microservice, the transfer will fail and be moved to the failed directory.
-   This will happen when the bag does not conform to the `BagIt`_ specification
-   or when one of the components is incorrect.
-
-#. Verify transfer compliance: if a job fails within this microservice, the
-   transfer will also fail and will be moved to the fail directory.
-
-#. Verify transfer checksums: if the checksums in the metadata
-   directory cannot be verified (i.e. if a file is missing or corrupted) this
-   microservice will fail and the transfer will be moved to the failed
-   directory.
-
 #. Scan for viruses: if a virus is found in the transfer then the microservice
    will fail and the transfer will be moved to the failed directory.
 
@@ -262,6 +246,20 @@ the transfer to be moved to the failed directory.
    the failed directory. The error logs will give details about the issue so
    you can investigate further.
 
+#. Approve Transfer (zipped and unzipped bags): When running a zipped or
+   unzipped bag transfer, the first microservice is 'Approve transfer.' If the
+   'verify bag, and restructure for complaince' job fails within this
+   microservice, the transfer will fail and be moved to the failed directory.
+   This will happen when the bag does not conform to the `BagIt`_ specification
+   or when one of the components is incorrect.
+
+#. Verify transfer checksums: if the checksums in the metadata
+   directory cannot be verified (i.e. if a file is missing or corrupted) this
+   microservice will fail and the transfer will be moved to the failed
+   directory.
+
+#. Verify transfer compliance: if a job fails within this microservice, the
+   transfer will also fail and will be moved to the fail directory.
 
 .. _common-errors:
 
@@ -294,7 +292,6 @@ workflow). This will move the transfer or SIP to the Rejected directory
 transfer or SIP will still be listed in the dashboard, however. See
 :ref:`Removing a transfer or SIP from the dashboard <removing>`, below, to
 remove it from the dashboard.
-
 
 .. _removing:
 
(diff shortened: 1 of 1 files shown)


getting-started/troubleshooting/error-handling.rst
```diff
@@ -228,18 +227,49 @@ a mistake was made entering the permalink:
...
-#. Verify metadata directory checksums: if the checksums in the metadata
-   directory cannot be verified (i.e. if a file is missing or corrupted) the
-   microservice will fail and the transfer will be moved in the failed
+There are a number of microservice errors that will halt the workflow and cause
+Archivematica to report a 'Failed transfer'. It is important to note that not
+all microservice errors will lead to a failed transfer. Below is a list of
+common errors that will result in the 'Failed transfer' microservice to run and
+the transfer to be moved to the failed directory.
+
+#. Approve Transfer (zipped and unzipped bags): When running a zipped or
```
sallain: This is a very minor thing, but an unordered list might be better than an ordered list here? I totally leave this up to you though.

getting-started/troubleshooting/error-handling.rst
```diff
@@ -228,18 +227,49 @@ a mistake was made entering the permalink:
...
+   the failed directory. The error logs will give details about the issue so
+   you can investigate further.
+
+
+.. _common-errors:
+
+Other common error behaviours
+-----------------------------
+
+Below is a list of common errors that, like normalization, will produce an
```
sallain, line 269:
```suggestion
Below is a list of common errors that will produce an
```
It could be confusing to refer to normalization here, if someone hasn't read the normalization section above.

getting-started/troubleshooting/error-handling.rst
```diff
@@ -283,20 +313,21 @@ in the dashboard:
    Click on the red Remove icon to remove a transfer or SIP from the dashboard,
    then click Confirm.
 
-It is recommended that you clear your dashboard of transfers and SIPs periodically
-to improve browser performance.
+It is recommended that you clear your dashboard of transfers and SIPs
```
sallain, line 313:
```suggestion
It is recommended that you :ref:`clean up the Transfer tab <cleanup>` and :ref:`the Ingest tab <cleanup-ingest>`
```

getting-started/troubleshooting/error-handling.rst
```diff
@@ -283,20 +313,21 @@ in the dashboard:
...
 Browser compatability
 ---------------------
 
-Archivematica has been tested most extensively with Firefox and Chrome. There are
-known issues with Internet Explorer 11 which result in an inability to start
-transfers in the dashboard (`Issue 7246`_).  Minimal, but successful,
-testing has been done with Microsoft Edge.
+Archivematica has been tested most extensively with Firefox and Chrome. There
+are known issues with Internet Explorer 11 which result in an inability to start
+transfers in the dashboard (`Issue 7246`_).  Minimal, but successful, testing
```
sallain, line 323:
```suggestion
transfers in the dashboard (`Issue 7246`_). Minimal, but successful, testing
```
Ooooh I thought I'd caught all the period+doublespace occurrences in the docs but every once in a while I find one!

sallain (changes requested):
A few small recommendations, but otherwise this looks good to me!

getting-started/troubleshooting/error-handling.rst
sarah-mason: good point!

getting-started/troubleshooting/error-handling.rst
sarah-mason: valid point. It saves future edits from making sure they remain in order.

sarah-mason:
don't mind the two commits. The first one wouldn't push to the branch and in the meantime I made another change. For some reason when I pushed that one it worked and both went through.

sallain (approved):

sallain:
@sarah-mason So that one Travis check is still failing, but it's because it's comparing against a specific, older commit - probably the commit that the 1.10 branch was at the last time you pushed. I've since fixed the TOC entry that was causing the error, but I don't know how to make Travis look at the correct commit. I think you should be good to merge - let me know if you aren't allowed.

ross-spencer:
Hi @sallain @sarah-mason to take into account changes to the TOC in upgrading.rst in a previous commit you need to rebase to bring this branch back in line to remove that warning.

You just need to be on this development branch locally, and then:
```
git fetch --all && git rebase origin/1.10
git push -f
```
And if you want before the git push, you can test with:
```
make html
```
Where it will complain beforehand, and not afterwards. 

This process will update the branch here and build correctly. I'm sure there's an updated version of [this](https://wiki.archivematica.org/Merging#Merge_process) somewhere? But it's step 3, replacing `qa/1.x` with `1.10` here because that's your target for the docs repo. 

I suspect you maybe right though Sara (trying to remember my git-fu), it might not matter when working on the individual file, you're not likely to overwrite anything, but that's the risk we're mitigating by following that step. 

**NB.** For Sarah M's sake if you haven't come across it, we try to never force push on a main branch as it will overwrite things irreversibly. We only do it on dev branches where we're trying to bring a dev branch in-line with main.

sallain:
@ross-spencer Thanks! @sarah-mason if you want to try following Ross' instructions, hopefully that will clear up that last pesky failure.

As a sidenote, I think the only time that the docs docs tell you to force-push is when you're updating a dev branch, which should be fine?

As a second sidenote, I always run `make clean html` rather than `make html` because the latter will only catch errors on files you've changed, whereas the former will check all files. Really helpful for a situation like this, where the error was caused by someone else merging with errors!

sarah-mason:
Thanks @sallain and @ross-spencer! Handy tips in there that I will try before I merge this morning. I knew yesterday when I ran `make clean html` I was still getting the error from the previous errors merged. I spent a fair bit of time reading about rebasing and testing various things to get it work properly (hence the two commits in a row), but I hadn't found these instructions! 🤞

sarah-mason:
Nice @ross-spencer -- that `git fetch` actually worked! So nice to finally see all green! 👍 

And thanks for all the explanations re. force pushing and what the instructions do. I seemed to have had a better time with rebasing by using `git fetch` rather than `git pull`. It may have had something to do with all the merges happening around this and the opaque nature of `git pull`.

PR #311, opened 2019-09-17
Update backlog and appraisal tabs

sallain:
This started out as a pretty simple change to the Backlog tab page, just intending to update some information... and then it spiralled out. Sorry about that!

I've substantially updated the information on both pages to be clearer and more accurate. I've also replaced screenshots throughout both backlog.rst and appraisal.rst and updated formatting to match the style guide where necessary.

PR #317, opened 2019-10-16
Fixes broken section links caused by PR #311

sallain:
I don't actually know why these broken links reappeared! They are definitely removed in #311 but somehow they're back. So let's remove them again.


sarah-mason (approved):
Let's hope it updates this time!

PR #328, opened 2019-11-12
remove link to known bug that has been fixed

pdurbin:
Ok to remove according to https://github.com/archivematica/Issues/issues/269#issuecomment-534687554


sallain (approved):
Thanks so much @pdurbin!

PR #343, opened 2020-01-17
Update DuraCloud integration page

sallain:
A general clean-up for the DuraCloud integration page.

After discussion with @ablwr and @dcraze, I have removed the sections
related to using DuraCloud as a transfer source location. Though it is
possible, early tests following the development of the integration
showed that using DuraCloud for this purpose was very slow and possibly
unreliable. Artefactual does not recommend this for Archives Direct
users, so it seemed better to remove the information. It can be added
back if that functionality improves.

This is connected to archivematica/Issues#171

### DIFF
--- a/admin-manual/installation-setup/integrations/duracloud-setup.rst
+++ b/admin-manual/installation-setup/integrations/duracloud-setup.rst
@@ -68,7 +68,7 @@ when you installed Archivematica.
    DuraCloud, you should create two spaces by repeating the instructions in Step
    1.
 
-#. In the Storage Service, navigated to the **Spaces** tab. You should see all
+#. In the Storage Service, navigate to the **Spaces** tab. You should see all
    of the DuraCloud spaces that you created in Steps 1 and 2.
 
 #. For each space, you need to create a corresponding :ref:`location
--- a/admin-manual/installation-setup/integrations/duracloud-setup.rst
+++ b/admin-manual/installation-setup/integrations/duracloud-setup.rst
@@ -45,11 +45,15 @@ DuraCloud can be used as a storage space for the following :ref:`purposes
 * **Transfer backlog**: A DuraCloud space used by Archivematica for material
   sent to the :ref:`backlog <backlog>` during processing.
 
+You can use one DuraCloud space for multiple purposes (e.g. one DuraCloud space
+could be used as both the AIP storage and transfer backlog), but we recommend
+managing these package types separately by creating DuraCloud spaces that are
+dedicated to the space's purpose. You do not have to create a space for all of
+these purposes, just those relevant your particular configuration. For AIP
+storage and DIP storage you can create multiple locations within DuraCloud.
+
 To add a space in DuraCloud, log in with administrator credentials and click on
-**Add Space**. You can give the space whatever name you wish. You do not have to
-create a space for all of these purposes, just those relevant your particular
-configuration. For AIP storage and DIP storage you can create multiple locations
-within DuraCloud.
+**Add Space**. You can give the space whatever name you wish.
 
 .. _duracloud-storage-service:
 
@@ -77,7 +81,7 @@ when you installed Archivematica.
 
    * **Purpose**: The purpose of the space (AIP storage, DIP storage, or
      transfer backlog.)
-   * **Pipeline**: Click on the pipeline name so that it is highlighted orange.
+   * **Pipeline**: Click on the pipeline name so that it is highlighted.
    * **Relative path**: The name of the space exactly as it appears in DuraCloud
      (e.g. ``aip-store``.)
    * **Description**: A human-readable description of the location. This will be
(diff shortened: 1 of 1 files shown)


admin-manual/installation-setup/integrations/duracloud-setup.rst
```diff
@@ -1,226 +1,107 @@
...
-   (e.g. "transfers") you will only need to configure the transfer source
-   once.
-
-.. tip::
-   If you browse for a path and do not see the expected directory listed, this
-   may be caused by a UI bug. You can type the path in instead.
+#. In the Storage Service's Spaces tab, click **Create new space** and enter the
+   following information:
+
+   * **Access protocol**: DuraCloud
```
ablwr: Should this section just link to the Storage Service docs instead? https://www.archivematica.org/en/docs/storage-service-0.15/administrators/#duracloud

ablwr: (or should this information be moved and updated there?, is the better question)

sallain: Good point @ablwr, thanks! I just went with what was already on the page but pointing out to the SS docs to reduce the number of places we need to keep updated is better.

ablwr (approved):

sallain:
@ross-spencer I've included an explicit reference to making sure that the space name doesn't include a leading slash.

admin-manual/installation-setup/integrations/duracloud-setup.rst
```diff
@@ -1,226 +1,101 @@
...
+#. In the Storage Service's Spaces tab, click **Create new space** and fill out
+   the form following the instructions in the :ref:`DuraCloud Storage Service
+   documentation <storageservice:duracloud>`.
+
+#. You must create a space for each storage location that you have created in
+   DuraCloud. For example, if you have AIP storage and DIP storage locations in
+   DuraCloud, you should create two spaces by repeating the instructions in Step
+   1.
+
+#. In the Storage Service, navigated to the **Spaces** tab. You should see all
```
ross-spencer:
```suggestion
#. In the Storage Service, navigate to the **Spaces** tab. You should see all
```

(or `once navigated to`)

admin-manual/installation-setup/integrations/duracloud-setup.rst
```diff
@@ -1,226 +1,101 @@
...
+<storageservice:location-purposes>` within Archivematica:
 
-Two DuraCloud users should be created:
+* **AIP storage**: A DuraCloud space where completed AIPs are placed for
+  long-term storage.
+* **DIP storage**: A DuraCloud space where DIPs can be stored. Note that it is
+  not necessary to store DIPs if you are sending them directly to an access
+  system.
+* **Transfer backlog**: A DuraCloud space used by Archivematica for material
+  sent to the :ref:`backlog <backlog>` during processing.
```
ross-spencer, line 47: I wonder if we can note that this is a recommendation, i.e. there are no technical reasons all three couldn't be `<a bucket>`. In doing this I feel it helps folk with the logic of what they're doing, as well as the logic of something that might go wrong e.g. if accidentally set up to be one bucket, then that's because it's all feasible. And we could add a lint of digipres and records and info mgmt too -- something along the lines of, management of these package types are best done separated into multiple bucket types dedicated to a storage service's purpose.

admin-manual/installation-setup/integrations/duracloud-setup.rst
```diff
@@ -1,226 +1,101 @@
...
+#. In the Storage Service, navigated to the **Spaces** tab. You should see all
+   of the DuraCloud spaces that you created in Steps 1 and 2.
+
+#. For each space, you need to create a corresponding :ref:`location
+   <storageservice:locations>`. Click on **Create Location here** to create a
+   storage location and enter the following information:
+
+   * **Purpose**: The purpose of the space (AIP storage, DIP storage, or
+     transfer backlog.)
+   * **Pipeline**: Click on the pipeline name so that it is highlighted orange.
```
ross-spencer: Accessibility: perhaps we can just say highlighted rather than orange? My own point of reference is that in my dark scheme I see largely in different colors, but other folk might have different barriers. (Also, in regular mode I didn't know this should be orange for me!) 

image

ross-spencer (commented):
This looks good Sara. I have left comments and suggestions but I also think technically it looks fine. 

What are your plans for the docs? This feels like a bit of a watershed where the storage service pages are more an appendix than anything? I do recognize that our choice to describe all space protocols on a single page there holds us back somewhat in terms of expression. Maybe we should bring them all together finally? (In time, in a managed way, etc. etc.)

admin-manual/installation-setup/integrations/duracloud-setup.rst
sallain: Excellent point!

admin-manual/installation-setup/integrations/duracloud-setup.rst
sallain: Thanks, I've included a note to that effect.

PR #354, opened 2020-02-10
Add section on special bag structures

sallain:
I've added a section about alternate configurations for bags. It's a bit
of a punt, as we don't test all of the possible bag configurations, but
I'm not sure how else to communicate that there may be other setup
options out there while also acknowledging that A) Artefactual doesn't
test them all, B) some of them might not work, and C) if they don't work
it doesn't mean it's a bug. So this is what I've landed on.

Connected to archivematica/Issues#669

### DIFF
--- a/user-manual/transfer/bags.rst
+++ b/user-manual/transfer/bags.rst
@@ -8,12 +8,16 @@ Archivematica supports the ingest of materials packaged in accordance with the
 Library of Congress `BagIt`_ specification. Users can ingest both zipped and
 unzipped bags by using the appropriate :ref:`transfer type <transfer-types>`.
 
-*On this page:*
+*On this page*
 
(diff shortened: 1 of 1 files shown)


ablwr (approved):

ross-spencer:
Late to the :tada: on this one, but to the extent I can help with testing of bags, and in future testing AMAUAT coverage, there's a [draft PR here](https://github.com/artefactual/archivematica-sampledata/pull/66) which brings in a custom processing config in one transfer, and a DemoTransferCSv type bag, with the paths updated to work correctly when processed. Anyway, hope it's helpful.

sallain:
Thanks @ross-spencer! Those are great to have.

ross-spencer:
I was hoping! And I meant to say as I introduced that PR, these are also good docs. Good to see them.

PR #355, opened 2020-02-20
Add basic information on transfer permissions

sallain:
This adds a section on the permissions that need to be given to the
Archivematica user for files and directories, and a note about what it
looks like when files/directories do not have sufficient permissions.

Connected to archivematica/Issues#1019

### DIFF
--- a/user-manual/transfer/transfer.rst
+++ b/user-manual/transfer/transfer.rst
@@ -15,6 +15,7 @@ Archivematica.
 
 * :ref:`Preparing digital objects for transfer <prepare-transfer>`
 
+  * :ref:`Transfer permissions <transfer-permissions>`
   * :ref:`Transfer types <transfer-types>`
   * :ref:`Basic transfers <basic-transfers>`
   * :ref:`Transfers with descriptive, rights, and/or event metadata <transfers-with-metadata>`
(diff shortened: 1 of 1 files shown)


user-manual/transfer/transfer.rst
```diff
@@ -126,6 +127,21 @@ Archivematica instance that you are using. For more information on configuring
...
 Archivematica <scaling-archivematica>`.
 
+.. _transfer-permissions:
+
+Permissions
+^^^^^^^^^^^
+
+In order for Archivematica to be able to carry out preservation tasks, the
+Archivematica user needs to have execute privileges for directories and read
+permissions for files in the transfer source location.
```
ross-spencer, line 137: This is good Sara. An obvious trip-up for a lot of folk. 

I wonder if it worth having some snippets here? I think Ops will have some shortcuts I don't but, could we go as far as to say, `0744` for  directories, `0644` for files (I think those are correct reading the instructions here), and show the `chmod` example? Maybe too `chown` for the AM user? 

It would be great to have for our reference too :) 

Did you ever see this [permissions calculator](http://permissions-calculator.org/) btw? I look it up all the time because I never remember the numbers, but maybe also useful to the reader here?

ross-spencer (commented):
Just left the one comment, this is a really great addition to have.

user-manual/transfer/transfer.rst
sallain: I originally wanted to add more specific instructions - but I spent about 90 minutes testing different combinations of permissions and there wasn't really a clear-cut minimum threshold of permissibility. So I decided to go broad!

ross-spencer: I hear you. We had that tricky ticket in 1.11 where it wasn't working at all, but it wasn't clear what was the right octal :) I wonder if a middle-ground might be to say, _YMMV but these work for us? e.g. on an ArchivesDirect install?_ - I see Miguel is penciled for review, so, maybe he can contribute to that? Otherwise, I appreciate the approach and going broad so that we have something and not nothing.

sromkey: I personally favour the broad approach, I suspect it makes enough sense to sys admins, and if it didn't for some reason they can reach out on the user forum, etc.

sromkey (approved):

PR #356, opened 2020-02-21
Update and reformat MediaConch instructions

sallain:
I've reformatted the MediaConch instructions so that the numbering renders
properly on the website. I've also updated the instructions for clarity
and used the example policy that is contained in the sampledata.

Connected to archivematica/Issues#780


ablwr (approved):

ablwr:
YESSS thank you!

PR #359, opened 2020-03-10
Update AIP structure page

sallain:
This page was a bit abandoned - it wasn't linked properly from the index
or any other page, so it had fallen out of date. I've done a
comprehensive update and added a lot more detail about the standard
structure of an AIP. I opted to use tree diagrams instead of screenshots
but maybe they're too fussy? Not sure.


ablwr (approved):

sarah-mason (approved):
I like the directory trees as examples. I really appreciate that you used a complex example (multiple transfers, one AIP). Nice update!

PR #364, opened 2020-03-30
Updates Elasticsearch information to add more info about consequences of disabling

sallain:
This information was a bit scatterd, so I've made it clear in several
places what the consequences are if you disable Elasticsearch. I've also
added a screenshot of the section of the admin tab that tells users the
Elasticsearch status and fixed a broken link.

Connected to archivematica/Issues#1158


ablwr:
You are so fast!!!!

ablwr (approved):

PR #365, opened 2020-03-31
Update metadata scripts for METS and PREMIS versions

sarah-mason:
I did not find any METS schema changes that were required, but some PREMIS Events and Rights metadata needed to be updated to version 3 from previous versions.

This is linked to issue: https://github.com/archivematica/Issues/issues/949


sallain (approved):
Ahh thanks so much @sarah-mason!

PR #372, opened 2020-04-15
Adds security note to README

sallain:
This is sample wording for a security note that we could add to all the repository READMEs.


sromkey (approved):
Looks good to me :+1:

PR #386, opened 2020-08-21
update ES index rebuild docs

peterVG:


admin-manual/maintenance/maintenance.rst
```diff
@@ -84,20 +86,64 @@ path of the AIP storage location you confirmed above.
...
 It can be executed multiple times with different paths to index multiple AIP
 storage locations.
 
+.. _aip-indexes-api:
+
+**Rebuild AIP indexes using Storage Service API**
+
+This command uses the Storage Service API to determine which stored AIPs and
+AICs need to be reindexed (based on status and origin pipeline), and then
+reindexes those AIPs and AICs from temporarily downloaded copies of the their
```
tw4l: I think "the" should be removed from "the their" at the end of this line

tw4l (approved):
I made one suggestion for a small change to make before merging but otherwise LGTM. Nice work, @peterVG !

admin-manual/maintenance/maintenance.rst
peterVG: Nice catch thanks. Change made. Merging...

PR #387, opened 2020-08-25
Update upgrading docs for Archivematica 1.12

tw4l:
- Remove 1.11-specific section on clearing watched directory
- Add section on updating Elasticsearch indices for 1.12

Connected to https://github.com/archivematica/Issues/issues/1257

### DIFF
--- a/admin-manual/installation-setup/upgrading/upgrading.rst
+++ b/admin-manual/installation-setup/upgrading/upgrading.rst
@@ -12,23 +12,29 @@ Upgrade from Archivematica |previous_version|.x to |release|
 * :ref:`Upgrade CentOS/Red Hat package install <upgrade-centos>`
 * :ref:`Upgrade in indexless mode <upgrade-indexless>`
 * :ref:`Upgrade with output capturing disabled <upgrade-no-output-capture>`
+* :ref:`Update search indices <update-search-indices>`
 
 .. note::
 
    While it is possible to upgrade a GitHub-based source install using ansible,
    these instructions do not cover that scenario.
 
+
 .. _completed-transfers:
 
 Clean up completed transfers watched directory
 ----------------------------------------------
 
-Upgrading from Archivematica 1.10.x or older to Archivematica 1.11.x can result
-in a number of completed transfers appearing as failed in the Archivematica
-dashboard, as well as corresponding failure notification emails being sent.
-These are not actual failures, but are unintentional side effects of changes
-made in Archivematica 1.11 to the workflow and to how metadata files are stored
-and copied into the SIP.
+.. note::
+
+   Ignore this section if you upgrading from Archivematica 1.11.
+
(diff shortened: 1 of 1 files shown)


admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -393,6 +364,112 @@ set environment variables and restart Archivematica processes.
...
+   locations may require use of the option to rebuild indices from the Storage
+   Service API rather than from the filesystem. At this time, it is not
+   possible to rebuild the indices for all types of remote locations.
+
+.. note::
+   Please note, the execution of this command may take a long time for big
+   AIP and Transfer Backlog storage locations, especially if the packages are
+   stored compressed or encrypted, or you are using a third party service. If
+   that is the case, you may want to consider using the
+   :ref:`Update the ElasticSearch mappings <update-mappings>` approach instead.
```
sevein: It should read "Elasticsearch" not "ElasticSearch".

admin-manual/installation-setup/upgrading/upgrading.rst
```diff
@@ -6,47 +6,18 @@ Upgrade from Archivematica |previous_version|.x to |release|
...
-   processing the transfers/ingests before proceeding.
-
-2. Delete all contents of the completedTransfers watched directory.
-
-   .. code:: bash
-
-      sudo rm -rf /var/archivematica/sharedDirectory/watchedDirectories/SIPCreation/completedTransfers/*
-
-3. Perform the upgrade as described below.
-
```
sevein: I think this is still relevant if you are upgrading from 1.10 or older, isn't it?

sevein (approved):
Super clear!

admin-manual/installation-setup/upgrading/upgrading.rst
tw4l: I was assuming that the instructions assumed users should do a stepped upgrade if jumping between versions, where they upgrade one version at a time. Not sure where I got that - maybe we used to recommend it? Happy to add it back if that's mistaken or we just want to err on the safe side.

tw4l:
Thanks @sevein ! Should be good for re-review now.

admin-manual/installation-setup/upgrading/upgrading.rst
sevein: I can't think of a reason why a stepped upgrade would be needed. @sromkey what do you think?

sevein (approved):

admin-manual/installation-setup/upgrading/upgrading.rst
sromkey: I have no idea! :P

tw4l: Hm, maybe I just made that up. Sounds like we should keep the completed transfers section in then, at least for a few versions!

PR #390, opened 2020-10-02
Apply migrations when CAS is enabled

sevein:
Describe how to apply django_cas_ng migrations when CAS is enabled.

Connects to https://github.com/archivematica/Issues/issues/1211.


replaceafill (approved):
:+1:

PR #392, opened 2020-12-02
Update instructions to use new single virtualenv location

sevein:
Connects to https://github.com/artefactual/archivematica/issues/573.

### DIFF
--- a/admin-manual/security/security.rst
+++ b/admin-manual/security/security.rst
@@ -282,7 +282,7 @@ CAS backend configuration in Archivematica Dashboard
             source /etc/sysconfig/archivematica-dashboard \
                 || (echo 'Environment file not found'; exit 1)
         cd /usr/share/archivematica/dashboard
-        /usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/python \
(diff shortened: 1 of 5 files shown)


admin-manual/security/security.rst
```diff
@@ -330,9 +330,7 @@ the three environments:
 
 .. code:: bash
 
-   $ sudo /usr/share/archivematica/virtualenvs/archivematica-dashboard/bin/pip install -U certifi
-   $ sudo /usr/share/archivematica/virtualenvs/archivematica-mcp-client/bin/pip install -U certifi
-   $ sudo /usr/share/archivematica/virtualenvs/archivematica-storage-service/bin/pip install -U certifi
+   $ sudo /usr/share/archivematica/virtualenvs/archivematica/bin/pip install -U certifi
```
replaceafill, line 331: The block above this still says `There are three virtual environments...`, should we update it too?

replaceafill (approved):
LGTM :+1:

admin-manual/security/security.rst
sevein: I will fix, thanks!

PR #400, opened 2021-02-02
remove hyphen

currmie:
eek, typo


ross-spencer (approved):
Both seemed like they could be right to me! Bowing to your judgement!

Issue #414, opened 2021-10-21
Problem: the Dublin Core field mapping used in AtoM DIP upload is not documented


peterVG:
Please do not file an issue here.  Instead, go to the [archivematica/issues](https://github.com/archivematica/Issues) repo to file your issue.

peterVG:
sorry, thanks @sevein :-)

peterVG:
i'll delete the PR branch too now

PR #415, opened 2021-10-21
add DIP Dublin Core to AtoM fields mapping

peterVG:
When AtoM receives a DIP upload from Archivematica it will look for Dublin Core metadata to populate archival description records. This documentation update adds a table that show how AtoM maps Dublin Core metadata to its fields, as per the source code here: https://github.com/artefactual/atom/blob/qa/2.x/lib/QubitMetsParser.class.php#L278-L366

Closes issue #414

### DIFF
--- a/user-manual/access/access.rst
+++ b/user-manual/access/access.rst
@@ -84,7 +84,9 @@ recognize by :ref:`using the arrangement feature on Appraisal tab
 
 You can also :ref:`add descriptive metadata <add-metadata>` to your transfer
 using either the metadata form or the metadata CSV file. This descriptive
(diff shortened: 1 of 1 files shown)


fiver-watson (approved):
Hi @peterVG , looks good! 

I'm approving, but I will include a couple ideas that could potentially extend the usefulness of them. First, might it be useful to include the ISAD standard field numbers? i.e. `title (ISAD 3.1.2)`, etc. Second, is there any commentary needed for the extent/format mapping? (i.e. how are the two appended to the same field? is there a carriage return between the values? does one always appear above the other? etc).

sromkey:
Thanks @peterVG , looks good, Dan's suggestions are good ones also but happy to have this merged as is. Technically should have been linked to an issue in the Issues repo as per the template instruction in this repo, but that's ok ;)

sromkey (approved):

PR #420, opened 2022-03-17
Update typo in DSpace Documentation

misilot:
Documentation was referencing Dataverse in the text for the url but is actually for DSpace.


sallain (approved):
Thank you for catching this, @misilot! Sorry that it took so long to review!

PR #426, opened 2023-09-19
Update manual install instructions for Ubuntu

marceloalencar:
As is, the install fails due to the missing MySQL Server.
Steps were updated according to the CentOS install guide and actual behavior during install.


replaceafill (approved):
Thank you so much @marceloalencar and apologies for the long delay getting back to you. I just tested your changes in a fresh Ubuntu 18.04 server and they worked great.

PR #440, opened 2024-02-20
add URLs to ignore

eviau-artefactual:
As part of https://github.com/archivematica/Issues/issues/708

PR #443, opened 2024-02-21
Fix a few github url to match the pattern described in the linkcheck configuration

eviau-artefactual:
Part of https://github.com/archivematica/Issues/issues/1638

To fit the pattern described in https://github.com/artefactual/archivematica-docs/pull/440

In other cases, I would have add a pattern to ignore the links that do not refers to .md files - right now I believe that this adds clarity to the URL? I might change my mind on this :sweat_smile:


replaceafill (approved):
Thanks. When you squash, please set the commit message and body as close as possible to https://github.com/artefactual/archivematica/blob/qa/1.x/CONTRIBUTING.md#commit-messages

PR #463, opened 2024-05-10
Improving linkcheck

eviau-artefactual:
Per https://github.com/sphinx-doc/sphinx/issues/8791#issuecomment-1895586426 , adding the `-q` or quiet option flag to the Sphinx options will allow us to only print out errors (broken URLs).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,9 +32,23 @@ jobs:
       - name: "Run build test"
         run: |
           make test
-      - name: "Linkcheck"
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v24
+      - name: Run linkcheck on .rst files
         run: |
-          make linkcheck
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            if [ "${file: -4}" == ".rst" ]
+              then
+                var="$var $file"
+            fi
+          done
+          if [ -z "$var" ]
+          then
+                echo "No *.rst changed files to check."
+          else
+                make files="$var" linkcheck-specified-files
+          fi
   lint:
     name: "Linting"
     runs-on: "ubuntu-22.04"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,10 +32,10 @@ jobs:
       - name: "Run build test"
         run: |
           make test
-      - name: Get changed files
-        id: changed-files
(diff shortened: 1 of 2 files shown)


eviau-artefactual:
Testing this: https://github.com/sphinx-doc/sphinx/issues/10643

eviau-artefactual:
It worked - with a few changes from the action set up in https://github.com/aiven/devportal/pull/1181

The license is https://github.com/aiven/devportal?tab=CC-BY-4.0-1-ov-file#readme - where/how should we make sure we do the attribution ?

eviau-artefactual:
Ok now it should work... but we will only know for sure when we make a change to a `.rst` file...

replaceafill (commented):
@eviau-artefactual After testing your branch a bit I realized that for my original pain point of `linkcheck` being too noisy I can instead use `make linkcheck SPHINXOPTS='-q'` which I think shows me only errors and it's supported by the current `Makefile`.

Also, don't you think that running the `linkcheck` target only on changed files could prevent us from detecting broken links early coming from unchanged files?

eviau-artefactual:
About the `-q` option: note that we are still using it, it's on line 192 in the Makefile - https://github.com/artefactual/archivematica-docs/pull/463/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R192 should be a direct link :)

If there were no external contributors to the documentation, e.g. if it was the work of a single person - it would make sense to linkhceck every link every time... The way I see this, is that right now, PRs can fail tests for things that are not related to the PR itself. If we can figure out a better way to do linkcheck, that is more granular, I think it would be best!

Maybe this conversation belongs in an issue ? If so, I can open one.

replaceafill:
> About the -q option: note that we are still using it, ...

I think my point is that we have always been able to set the `-q` flag for the existing `linkcheck` target. I just didn't know how to do it using `SPHINXOPTS`.

> The way I see this, is that right now, PRs can fail tests for things that are not related to the PR itself.

Maybe that's what we need to fix.

> Maybe this conversation belongs in an issue ? If so, I can open one.

Definitely and thank you! I think we've tried to address my pain points and yours in a single PR :sweat_smile:

eviau-artefactual:
Issue 1690 reference the verbose results and makes them quiet.

Issue 1689 will follow through on the implementation of linkcheck.

Closing this PR.

PR #465, opened 2024-06-27
Add instructions to build local, offline docs

eviau-artefactual:
Part of https://github.com/archivematica/Issues/issues/1701

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,11 @@ to the Internet.
 While this offline version will not have the Archivematica web theme, you will
 gain access to improved search features.
 
+This local version will contain only the Archivematica documentation; notably
+the Archivematica Storage Service docs will be missing. Instructions on how to
+build a local, offline version of the Archivematica Storage Service is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,68 @@ You are free to copy, modify, and distribute the Archivematica documentation
...
+open _build/html/index.html
+```
+
+The HTML files for the documentation will be in
+`archivematica-docs/_build/html/`.
+You can open the files in a browser of your choice, without having any access
+to the Internet.
+
+While this offline version will not have the Archivematica web theme, you will
+gain access to improved search features.
```
replaceafill, line 75: Do you think this should also point or warn that the Archivematica Storage Service docs won't be locally available?

replaceafill (approved):
LGTM.

README.md
eviau-artefactual: I could add something about this, yes - and I could also make a PR for offline, local version of the Archivematica Storage Service docs.

PR #466, opened 2024-07-02
Ignore website for linkcheck

eviau-artefactual:
Found during https://github.com/artefactual/archivematica-docs/pull/465

I think we can ignore this as I believe we use this URL as an example.com placeholder... Let me know if that is not the case.


replaceafill (approved):
Thanks!

PR #468, opened 2024-07-10
Tentative improved linkcheck

eviau-artefactual:
Hypothesis:

- we want quiet results : https://github.com/archivematica/Issues/issues/1689
- we want to run a linkcheck regularly, but not for every PR: https://github.com/archivematica/Issues/issues/1690

Proposal:
- to run a monthly linkcheck, that we can also run manually if needed.

Thoughts ? Considerations ?


eviau-artefactual:
I am going to close this so it has three parts:

#472 
#473 

and an upcoming PR with the scheduled job.

See https://github.com/archivematica/Issues/issues/1690#issuecomment-2271340072 for discussion.

PR #469, opened 2024-07-15
Enable full search features with HTTP server

eviau-artefactual:
You will need the server to access full search features, such as highlighting the context surrounding keywords you are looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ sphinx-build -D language=pt_BR ./ _build/html/pt_BR # for *português do Brasil*
 * Access the documentation:
 
 ```bash
-python3 -m http.server --d _build/html
+python3 -m http.server -d _build/html
 ```
 
 The HTML files for the documentation will be in
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ sphinx-build -D language=pt_BR ./ _build/html/pt_BR # for *português do Brasil*
 * Access the documentation:
 
 ```bash
-open _build/html/index.html
+python3 -m http.server
 ```
 
 The HTML files for the documentation will be in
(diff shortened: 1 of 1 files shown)


replaceafill (commented):
With this approach, wouldn't you need to explain users that they have to point their browser to `http://localhost:8000` or something? And maybe also use the `--directory/--d` flag to point to the `_build/html` directory directly?

eviau-artefactual:
I did not know about that flag! Thanks!

eviau-artefactual:
The linkcheck test fails because of:

```
(getting-started/overview/external-tools: line    7) broken    https://www.gnu.org/licenses/agpl-3.0.en.html - HTTPSConnectionPool(host='www.gnu.org', port=443): Max retries exceeded with url: /licenses/agpl-3.0.en.html (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1007)')))
```
We could review this later on. Adding this as an example of https://github.com/archivematica/Issues/issues/1690

README.md
```diff
@@ -63,13 +63,13 @@ sphinx-build -D language=pt_BR ./ _build/html/pt_BR # for *português do Brasil*
 * Access the documentation:
 
 ```bash
-open _build/html/index.html
+python3 -m http.server --d _build/html
```
replaceafill: Apologies. I suggested a typo in my previous comment. Even though it works I think using `-d` is more correct.

```suggestion
python3 -m http.server -d _build/html
```

replaceafill (approved):
:+1:

PR #471, opened 2024-08-06
Remove outdated contract part - FAQ

eviau-artefactual:
We noticed the FAQ needs an update about the contracting part.


sromkey (approved):
lgtm!

PR #473, opened 2024-08-06
Make linkcheck quiet

### LINKED ISSUE #1689: Problem: the verbose results that we got from linkcheck makes it difficult to pinpoint which URL needs an update
## Current behaviour

At the moment speaking, running `make linkcheck` gives us a long list of URLs, most of which are labeled as working, some of which are broken.

It would be nice to only output URLs that need to be updated.

## Potential solution

Use the `-q` option flag for sphinx to get *quiet* results: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-q

## Other relevant information

The original PR for this was https://github.com/artefactual/archivematica-docs/pull/463 - I am splitting the two concerns in two issues to better address each one.

eviau-artefactual:
Closes https://github.com/archivematica/Issues/issues/1689

Could also be as a variable - let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ changes:
 
 .PHONY: linkcheck
 linkcheck:
-	$(SPHINXBUILD) -b linkcheck $(ALLSPHINXOPTS) $(BUILDDIR)/linkcheck
+	$(SPHINXBUILD) -b -q linkcheck $(ALLSPHINXOPTS) $(BUILDDIR)/linkcheck
 	@echo
 	@echo "Link check complete; look for any errors in the above output " \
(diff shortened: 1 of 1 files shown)


replaceafill:
What do you think of leaving the `Makefile` as it is?

[People like me](https://github.com/artefactual/archivematica-docs/pull/463#issuecomment-2112898448) who want to focus only on errors can rely on the existing `SPHINXOPTS` variable already:

```
make linkcheck SPHINXOPTS='-n -q'
```

eviau-artefactual:
Works for me! I will close this PR.

PR #474, opened 2024-08-13
Pin Sphinx version (temporary)

eviau-artefactual:
This should allow the docs to pass the build test.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-sphinx==7.4
+sphinx==7.4.7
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1 +1 @@
-sphinx
+sphinx==7.4
```
replaceafill:
```suggestion
sphinx==7.4.7
```

replaceafill (approved):

PR #478, opened 2024-09-25
Document v2beta package endpoint behavior

replaceafill:
It states what the endpoint does when the transfer source location's UUID is not explicitly passed and how to use it when multiple locations are available.

Connected to the first point of https://github.com/archivematica/Issues/issues/1709#issuecomment-2294186837


sromkey (approved):
👍

PR #497, opened 2025-12-10
Document new cookie and session security variables

replaceafill:
This updates the cookie and session security section to include the
Dashboard application variables and the Storage Service
application-specific environment variables introduced in 1.18.0.

This is how the section looks now:
image


mamedin (approved):
Thanks!

